PR #7, opened 2014-11-05
Don't fail if Asterisk and asterisk_click2dial aren't installed

bwrsandman:
When adding telephony to Odoo's addons path, it will fail if Asterisk isn't installed.
Since Asterisk is declared as an external dependency, it cannot be installed in this case.
However, odoo will still try to execute code imported in `__init__.py` even if the module isn't installed.

Hiding the ImportError is important for users who do intend to install asterisk_click2dial but need the modules included in this repo.


pedrobaeza:
:+1:

alexis-via:
:+1:

alexis-via:
I also cherry-picked it in the 8.0 branch

Issue #11, opened 2014-12-31
Test connection is always announced as successful


Garamotte:
On Odoo v8.0, when using the test button on Asterisk server configuration, the success message is displayed, even if the test failed.

I added a "print" instruction just before the error message (Connection Test Failed") and the string is printed, with the success message displayed on Odoo.
I think this is because the success message is sent from the "finally" block.

PR #12, opened 2015-01-02
[FIX] Bug #11 AMI test: success is now only reported when it really works

alexis-via:
I wonder who is the stupid developper that introduced bug #11 !!! :)
It is now fixed.


pedrobaeza:
:+1:

alexis-via:
we really need to merge this :+1:

pedrobaeza:
Merging...

PR #13, opened 2015-01-02
FIX: clear field when number not set rather than leaving previous value

tbhi:
Fix for switching between contact via form view showing the previous contact's phone number if this contact does not have one set.


alexis-via:
Thank stan3 for fixing this very important bug !
Code review and test with base_phone alone and base_phone + asterisk_click2dial with a real test of click2dial :+1:

bguillot:
:+1: 

Thanks

yvaucher:
Ok travis error are unrelated I proceed with merge.
Thanks!

Issue #17, opened 2015-01-23
base_phone : we should not raise in create() or write()


alexis-via:
With the current code of base_phone, if the phone number is very bad and cannot be reformatted, it will do a raise (cf base_phone/base_phone.py line 101). This is good when the create() or write() is made by a user in the Web interface : that way, the user is aware that there is a problem in the phone number ; he can fix this issue and create/write again.

But, when the create() or write() is made by an external system (for example, the partner is created by the Amazon connector), then it's not a good thing to raise in the creation of the partner, because it will block the import of the order !

The problem is that I don't think we are able to differentiate inside the create() or write() is the call is made by a real user in the web interface (in this case, we would do a raise) or if the call is made "by code" (in this case, we would just log the problem).

So, if we remove the raise, how could we warn the user in the web interface about the fact that the phone number is wrong ?

alexis-via:
PR has been merged

PR #18, opened 2015-01-23
Fix bug #17: don't raise by default on write/create

alexis-via:
As a consequence, we can have non E.164 values inside the phone/mobile/fax fields: in this case, we don't display an href link on the field and we don't display the Dial button.

Fix show old number on paging for fax field (the fix in PR  #13 only fixed the issue for the phone field, but the bug was also present in the fax field !)

### DIFF
--- a/asterisk_click2dial/asterisk_click2dial.py
+++ b/asterisk_click2dial/asterisk_click2dial.py
@@ -347,7 +347,8 @@ class res_users(orm.Model):
             help="Caller ID used for the calls initiated by this user."),
         # You'd probably think: Asterisk should reuse the callerID of sip.conf!
         # But it cannot, cf
-        # http://lists.digium.com/pipermail/asterisk-users/2012-January/269787.html
+        # http://lists.digium.com/pipermail/asterisk-users/
+        # 2012-January/269787.html
         'cdraccount': fields.char(
             'CDR Account', size=50,
             help="Call Detail Record (CDR) account used for billing this "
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Don't you change the widget to add `raise_if_parse_fails` argument in interactive mode?

alexis-via:
@pedrobaeza Yes, that's a very good idea. I don't know how to do that exactly.
Maybe a simple solution would be to do in the XML view:

```
        <field name="phone" position="attributes">
            <attribute name="widget">phone</attribute>
            <attribute name="context">{'raise_if_parse_fails': True}</attribute>
        </field>
```

and then read the context to get the raise_if_parse_fails in the inherit of write/create.

Did you have another idea ?

pedrobaeza:
Yes, that is the only way I see, because you can't control it in JS. Let's do it via context, but you can keep the parameter, and look for context key on create/write and pass it as an argument to the method.

alexis-via:
@pedrobaeza The solution I proposed in my previous comment didn't work (the context inside the "phone" field wasn't passed to create/write), but I added the context in the window action, and it works.

If we think this is the right solution, I would need to inherit the other window actions that contain phone numbers.

pedrobaeza:
Yeah, I see. That's why I told about putting in the widget, but the reformatting occurs when saving, so it's a different flow.

If no one tells about another solution, we would have to do it that way.

chafique-delli:
It's ok for me. :+1:

alexis-via:
I use this branch in production on several odoo v8 systems :+1:

PR #23, opened 2015-02-21
Update screenshots of asterisk_click2dial

alexis-via:
Remove old-style icons on buttons


pedrobaeza:
You have seen @StefanRijnhart 's comments :wink: 

:+1:

StefanRijnhart:
LOL, thanks! :+1: of course although I must say that I had a little bit of trouble trying to discover what was new in the partner form view (of course, it was the 'dial' link). So I would maybe try and highlight this a little bit setting the contrast of this region a little lighter in an image editor (or the rest a little darker).

base_phone/wizard/number_not_found_view.xml
```diff
@@ -39,7 +39,7 @@
                         options="{'dial_button_invisible': True}"/>
                     <field name="current_partner_mobile" widget="phone"
                         options="{'dial_button_invisible': True}"/>
-                    <button name="update_partner" icon="gtk-convert" colspan="2"
+                    <button name="update_partner" class="oe_highlight" colspan="2"
                         string="Update Partner with this Number" type="object"/>
```
yvaucher, line 43: Wouldn't it be better to put those buttons in `<footer>` ?

Using visibility to show `Update` if a partner has been selected `Create new partner` if no partner was selected.
Plus add a text before `partner_id` saying something like "Select a partner to update it"

I did something like this in:
https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_shipment_management/wizard/create_shipment_view.xml#L8-36

alexis-via, line 43: @yvaucher your proposal is certainly a good idea. We'll try to do that in another PR.

PR #34, opened 2015-03-25
FIX old/new API call

alexis-via:
Obvious fix, sorry for that. I am so used to code with new API now... !


alexis-via:
Forgot to mention that I made a real-life click2dial with my home asterisk server and it works well.

pedrobaeza:
:+1:

alexis-via:
OK, let's fast-track this.

PR #38, opened 2015-04-09
edit arguments accordingly to func definition

igor-shtein:
small addition to 62cb85c


alexis-via:
Ooops, it's my fault. Thanks for this fix :+1:

sebastienbeau:
Thanks for the fix :+1:

Issue #40, opened 2015-04-14
Inbound number that does not match a partner results in error.


jaredkipe:
This may be a configuration issue, but I haven't been able to find a place that would help.

```
Traceback (most recent call last):
  File "/share/odoo/openerp/http.py", line 530, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/share/odoo/openerp/http.py", line 567, in dispatch
    result = self._call_function(**self.params)
  File "/share/odoo/openerp/http.py", line 303, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/share/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, *args, **kwargs)
  File "/share/odoo/openerp/http.py", line 300, in checked_call
    return self.endpoint(*a, **kw)
  File "/share/odoo/openerp/http.py", line 796, in __call__
    return self.method(*args, **kw)
  File "/share/odoo/openerp/http.py", line 396, in response_wrap
    response = f(*args, **kw)
  File "/share/odoo/addons/web/controllers/main.py", line 936, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/share/odoo/addons/web/controllers/main.py", line 928, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/share/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/share/scg-odoo-community/connector-telephony/base_phone/wizard/number_not_found.py", line 64, in default_get
    cr, uid, None, {'phone': res.get('calling_number')}, context=context)
  File "/share/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/share/scg-odoo-community/connector-telephony/base_phone/base_phone.py", line 38, in _generic_reformat_phonenumbers
    assert isinstance(self._country_field, (str, unicode, type(None))),\
AttributeError: 'phone.common' object has no attribute '_country_field'
```

jaredkipe:
I added another assert above the one that was failing like:

``` python
assert hasattr(self, '_country_field'), 'Object missing _country_field for object "' + str(self) + '"'
```

which gives a similar stacktrace but with the following

```
AssertionError: Object missing _country_field for object "phone.common"
```

Later down this file, there is an extension to 'res.partner' that mentions phone.common...

jaredkipe:
[...]

PR #41, opened 2015-04-14
8.0 Fix bug #40

alexis-via:
This patch is untested for the moment.


jaredkipe:
+1

jaredkipe:
+1 for changes to requirements (e.g. adding bus_enhanced)  I was about to make a request for this.

alexis-via:
Travis is red, but it is caused by a bug in Odoo that I reported https://github.com/odoo/odoo/issues/6324 so it's not our fault.

So I think we are ready to merge this fix thanks to the tests of @jaredkipe  :+1:

PR #46, opened 2015-05-13
8.0 FIX a crash when exception has no len

alexis-via:
I am not very good at exception handling, so I don't know if my patch is the right solution to the pb.


alexis-via:
Crash needs to be merged :+1: 
If someone has a better patch, we can merge it later.

PR #49, opened 2015-06-18
use missed opportunity to detect number

igor-shtein:
add 'ringing' state number detection on new asterisks versions

### DIFF
--- a/asterisk_click2dial/asterisk_click2dial.py
+++ b/asterisk_click2dial/asterisk_click2dial.py
@@ -253,6 +253,13 @@ class asterisk_server(orm.Model):
                     _logger.debug("Found a matching Event in 'Ring' state")
                     calling_party_number = chan.get('CallerIDNum')
                     break
+                # Compatibility with Asterisk 1.4
+                if (
+                        chan.get('State') == 'Ringing' and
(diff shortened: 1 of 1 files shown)


alexis-via:
Code review, no test (Asterisk 1.4 is soooo old, how could I test that :)
:+1:

asterisk_click2dial/asterisk_click2dial.py
```diff
@@ -253,6 +253,13 @@ def _get_calling_number(self, cr, uid, context=None):
                     _logger.debug("Found a matching Event in 'Ring' state")
                     calling_party_number = chan.get('CallerIDNum')
                     break
+                # Compatibility with Asterisk 1.4
+                if (
```
pedrobaeza, line 257: Why this strange line break?

igor-shtein, line 257: I actually copy-paste "if" block from below, with comment and line breaks I assume.
I left comment about version coz it seems to be same api.
I`m running version 11.7 of asterisk and this change need to be in order to detect number in ringing state.
About line break, I used github built-in editor. I sure can reedit this with normal editor.

pedrobaeza:
This can be merged although old.

Issue #20, opened 2015-02-05
Asterisk calling previously failed calls


aaltinisik:
Thank you for the module :)

if we make some calls and hangup before connection or do not pickup the phone, 
on next click2call action old number is dialed instead of new one.

So it remembers failed numbers and try to connect you. Is this from asterisk or connector ?

alexis-via:
You should analyse the trafic between asterisk and port (tcp port 5038 by default), and you will know if the problem is on the asterisk side or on the openerp side:

hint:
tcpdump -n -i eth0 -A -s 0 port 5038

aaltinisik:
I am pretty sure it is from openerp side. (from pyAsterisk library) 
directly connected to AMI instead of pyasterisk and no problem.

aaltinisik:
but cannot pinpoint the problem,

agutierrezdapda:
can you test with this code. Work for me.:

```
+++ base_phone/static/src/js/phone_widget.js
@@ -18,6 +18,8 @@
             if (!this.get('effective_readonly')) {
                 this._super();
             } else {
+                // cleanup
+                this.$el.find('#click2dial').off('click');
                 var self = this;
                 var phone_num = this.get('value');
                 //console.log('BASE_PHONE phone_num = %s', phone_num);
```

alexis-via:
@agutierrezdapda I made a PR with your patch and quoted your name in the commit (well, I don't know your real name, so I quoted your github account). Could you make a "+1" on the comment of the PR #52

intero-chz:
Hey guys,

it seems there is the same problem on branch 7.0. Maybe the patch is fixing the problem there, too. I will let you know and create a PR for 7.0 then.

best regards

sadasant:
This also happens in the version available for Odoo 10, and the solution is similar to what has been mentioned: https://github.com/OCA/connector-telephony/pull/157

PR #53, opened 2015-06-25
Fix bug #20 Multiple click2dial when previous click2dial failed

alexis-via:


alexis-via:
The previous identical PR has already been approved, so I merge it.

PR #58, opened 2015-09-28
Freeswitch 7.0

treveradams:
This is 7.0 with freeswitch_click2dial and freeswitch_click2dial_crm. These are fully tested with OpenERP 7.0. I have updated the translations as I could (mostly regex).


treveradams:
The remaining errors seem to be some formatting complaints, the fact that ESL module is only available if you build it as part of a FreeSWITCH installation, and complaining that geocoder isn't used when it is.

PR #60, opened 2015-09-28
Override name_get for hr_applicant so that applicant name for cid lookup is returned

treveradams:
Override name_get for hr_applicant so that applicant name and not position is what is returned by caller ID lookup.

Do not return any information if applicant is hired or refused as this information is either useless or exists elsewhere.

Possible enhancement:
- Create three global settings: allow cid for hired applicant, allow cid for refused applicant, show title and not applicant name for cid.


treveradams:
It appears that all of these errors are a problem with travis, not the code as they are modules that should be installed.

treveradams:
That and in_group_20 in the asterisk click2dial module, which is not modified here.

PR #61, opened 2015-09-28
Override name_get for hr_applicant so that applicant name for cid lookup is returned

treveradams:
Override name_get for hr_applicant so that applicant name and not position is what is returned by caller ID lookup.
Do not return any information if applicant is hired or refused as this information is either useless or exists elsewhere.

Possible enhancement:
- Create three global settings: allow cid for hired applicant, allow cid for refused applicant, show title and not applicant name for cid.

### DIFF
--- a/hr_recruitment_phone/hr_recruitment_phone.py
+++ b/hr_recruitment_phone/hr_recruitment_phone.py
@@ -42,3 +42,19 @@ class hr_applicant(orm.Model):
             cr, uid, ids, vals, context=context)
         return super(hr_applicant, self).write(
             cr, uid, ids, vals_reformated, context=context)
+
+    def name_get(self, cr, uid, ids, context=None):
+        if context is None:
+            context = {}
+        if context.get('callerid'):
+            res = []
+            if isinstance(ids, (int, long)):
+                ids = [ids]
+            for applicant in self.browse(cr, uid, ids, context=context):
+                name = applicant.partner_name
+                if applicant.state not in ['done', 'cancel']:
+                    res.append((applicant.id, name))
+            return res
+        else:
+            return super(hr_applicant, self).name_get(
+                cr, uid, ids, context=context)
--- a/hr_recruitment_phone/hr_recruitment_phone.py
+++ b/hr_recruitment_phone/hr_recruitment_phone.py
@@ -43,6 +43,20 @@ class hr_applicant(orm.Model):
         return super(hr_applicant, self).write(
             cr, uid, ids, vals_reformated, context=context)
 
+    # The only reason this is overloaded her as applicant.name is the
+    # position. This is what base_phone.py:name_get will return.
+    # It makes more sense in this context to return the applicant's name.
+    # This code appears weird as it returns [] in the cases:
+    # 1) hired - done or 2) refused - cancel.
+    # This is done so that an employees record is used instead of the
+    # applicant record. The following (cancel) applies here as well.
+    # It is done for cancel so that an applicant who is refused doesn't
+    # cause problems if that number is reassigned to another entity in the
+    # future or the same person applies for the same or different position
+    # in the future. We cannot call super name_get in these cases as it will
+    # still return the position the person applied for and the pop-up
+    # functionality would then pull up this record. Both of these are
+    # undesirable.
     def name_get(self, cr, uid, ids, context=None):
         if context is None:
             context = {}


hr_recruitment_phone/hr_recruitment_phone.py
```diff
@@ -42,3 +42,19 @@ def write(self, cr, uid, ids, vals, context=None):
...
+        if context is None:
+            context = {}
+        if context.get('callerid'):
+            res = []
+            if isinstance(ids, (int, long)):
+                ids = [ids]
+            for applicant in self.browse(cr, uid, ids, context=context):
+                name = applicant.partner_name
+                if applicant.state not in ['done', 'cancel']:
+                    res.append((applicant.id, name))
```
alexis-via, line 70: When applicant.state is done or cancel, you return []... which is a bit wired !

alexis-via:
When my remark on the code will be taken into account: :+1:

treveradams:
I believe my patch is accurate as it mirrors what the base_phone.py (line 250-259 in my local copy) module does when it can't find the data. I also believe the following actual test results (comment before colon, data in quotes after) show empirically that it is correct. Please, correct me if I am wrong. I added employee and applicant to the records in question to show what is going on. It is actual results, not me adding things in the results.

The following is a URL to access a script to fetch the caller/callee information in the proposed FreeSWITCH module.

http://openerp.example.org:8080/wscgi-bin/get_caller_name.py?name=unknown&number=9812398524&notify=1005 (this is in my FreeSWITCH tree and is the way to fetch the information for caller and callee setting, with notify to internal extension 1005)

Test data being used (Shane Williams, applicant for "Programmer")

Without my patch:
No matter the stage (cancel, done, etc.) of Shane Williams: "Programmer"
I didn't retest this, but I believe this problem exists even if you then hire Shane... the applicant record is what is returned.

With my patch:
Status refused: "unknown"
Not refused or hired: "Shane Williams, applicant"
Hired with name changed (in employee record) to show it is working: "Shane Williams, employee"
Delete employee, move Shane back to initial: "Shane Williams, applicant"
And back to refused: "unknown"

If the refused is somehow not accurate, I suppose the most correct way would be to change to the following:

```
def name_get(self, cr, uid, ids, context=None):
    if context is None:
        context = {}
    if context.get('callerid'):
        res = []
        if isinstance(ids, (int, long)):
            ids = [ids]
        for applicant in self.browse(cr, uid, ids, context=context):
            name = applicant.partner_name
            if applicant.state not in ['done', 'cancel']:
                res.append((applicant.id, name))
            else
                res.append(super(hr_applicant, self).name_get(
            cr, uid, ids, context=context))
        return res
    else:
        return super(hr_applicant, self).name_get(
            cr, uid, ids, context=context)
```

However, I am not sure res.append will work that way.

treveradams:
alexis-via: I have rechecked my assumptions based on the original code in base_phone. I know my possible fix posted above is incorrect. I also still believe my code is correct (functionally and otherwise). Please, review my code compared to base_phone. If I am incorrect I will look at it again and try to figure out why it is different and what the correct way is.

Thank you.

treveradams:
Alexis, is it that I am hiding an applicant if he/she has been hired (done) or refused (cancel) that is the problem and not so much the style? I did this so that hired employees are found by their record as an employee and not as the applicant. I did it for refused as the applicant may apply again with updated information, for a new position, etc. That will be the record one wants to come up, not an old one. Additionally, in both cases, people may get new phone numbers as people release old ones (while this is less common in the US than it was, it still happens, my cell phone number is an example as it used to belong to a radio personality and for years I got calls from his friends/family). Perhaps that doesn't happen elsewhere (I am not familiar with telephone laws/practices elsewhere... I imagine people do at least occasionally move from city to city, etc.).

In all of these cases (hired - done, refused - cancel) it seemed the right thing to do with records that were not in use any long to not be used to provide caller ID name information. Any attempt to call the super name_get (if I remember correctly) results in the position being returned instead of applicant name in the case where there was an applicant and they were hired/refused instead of ignoring the record. Without it, as the test case (again, I didn't just tell you what it should do, I typed up things as I tested them) shows it functions as it should. I didn't test with partners (non-employees, etc.), but I believe the results should be the same.

The unknown is provided by the input as shown to the script and not by my code in OpenERP/Odoo.

So, those were my design decisions and led to my suggestion of the three suggested enhancement options (which I had not successfully implemented yet).

treveradams:
@alexis-via would you mind telling me what you think it should return or if my supposition is correct since it is what the base function does?

PR #62, opened 2015-09-29
[FIX] set countrycode in upper case to avoid error with the lib phonenumbers

bguillot:
Hello, 

The library phonnumbers only accepts upper case country code. Set the countrycode variable in upper case can avoid some problems when reformating.


pedrobaeza:
:+1:

alexis-via:
:+1:

PR #63, opened 2015-09-30
ported to Odoo 9

jmalinens:
Did a little bit of work to port it to Odoo 9.

Some suggestions:
- switch to JSON-RPC instead of XML-RPC
- adding web_live, web_live_kanban, web_action_request, live_crm modules to the list of modules. I have also ported those. Do You need a pull request for that?
- how about integrating JsSIP in the future?

Cheers,
Juris Malinens from Quinto Networks


pedrobaeza:
Is this compatible with both v8/v9?

alexis-via:
@jmalinens It seems you try to push this to the wrong branch. The 9.0 branch is not created yet.

jmalinens:
I don't think so. Javascript has changed quit a lot and also some views have changed. Models should work and could be backported to 8.0 (just because new API is faster) but I don't have time myself to backport it. But I kept same functionality and haven't added/deleted any functionality so it should work the same as in 8.0. I know it is not created that's why I created it myself ;)

pedrobaeza:
OK, then you have to wait a little until 9.0 is finally out, and OCA create the 9.0 branch for this repo. Then, check that all continue working (maybe last changes affect), and make again the PR against this 9.0 branch.

jmalinens:
Sure!

PR #65, opened 2015-10-09
BaseFax and SFax Adapter

lasley:
This PR add base_fax and an adapter to use SFax. It's basically a rough outline at this point (although everything works great). 

I wanted to maybe get some opinions on how I am structuring this before I go any further. I have kind of a weird inheritance system similar to what you would see in the message library in order to centralize alot of the base methods, while still allowing for proprietary adapters fitting the interface. I think maybe I could have used an AbstractModel here, but I got incredibly confused and ended with this.

Because of this funky inheritance, I ran into an issue when trying to present the SFax adapter in views with the Base interface. I ended up putting the SFax adapters in Company for the time being, but I think someone more experienced may have a better idea here?

The only features in this request are Fax adapters, payloads, tranmissions, and send via SFax. 

I will be adding Status callbacks, and Inbound fax as well, I just wanted opinions before I got too deep down this rabbit hole. Please advise if this was not the proper way to gather those opinions :wink:


lasley:
Added inbound Sfax callbacks and fixed some of the view/inheritance issues I was talking about. Also, everything else is more streamlined in general.

Setting up the adapters is still kind of clunky, but at least it's not in two spots anymore. 

Faxes and Odoo are kind of an interesting combination of old :puke: and new :smile:

treveradams:
Just an idea that I was playing with for some ideas with FreeSWITCH faxing: It may be good to allow the system to keep track of retries in this way.

1) Fax system attempts to send, gets out three confirmed pages
2) System retries the send using some new cover page, but renumbers pages, after first page it starts with the first failed page

So, can your system accommodate this with appropriate logging? Or do you think the only logging attempt number is enough?

treveradams:
Additionally, as I dig through this, I see you do not have a callback (XML-RCP? JSON-RPC?) where a faxing service could notify you of an received fax and the id.

treveradams:
Or, more likely upload the data directly through the RPC interface as a PDF. This is what I was thinking of doing for my system. Maybe both methods would be good for systems that can/will directly send to Odoo and for Odoo systems that think this is an acceptable way (such as a local FreeSWITCH or Asterisk system).

lasley:
Hi @treveradams - Thanks for the comments!

There are callbacks implemented in the fax_sfax controller that handles both inbound and outbound statuses. Are you thinking that I should instead centralize this? I'm not sure how the other fax service callbacks work, so I wasn't really sure what aspects of that controller could be abstracted.

As you see, all I'm doing is logging the amount of attempts and final error message, but there is not a scheduled task governing the logging. It seems that the callback is working pretty well for this purpose, although there is the possibility of not receiving a status due to system failure.

I'm not too sure how to leverage the RPC interface with the callbacks though. Sfax in particular is sending the callbacks via HTML application type, not JSON, so I ran into issues using standard protocols. Sounds like you have some ideas in this regard?

lasley:
Btw fax download is triggered by the callback using `fax.adapter.sfax._fetch_payloads`

treveradams:
Please consider https://github.com/laslabs/connector-telephony/blob/e83c3632d85cf1fb502e059004b2e7caf7cb37a6/base_phone_popup/popup.py:incall_notify_by_login. What I am thinking is a fax system such as FreeSWITCH could call the RPC with the phone number (if you look at my code for this you will see an incall_notify_by_extension) and the fax PDF, inbound and outbound numbers, etc. and the fax would be stored. It should only be visible to the user with the extension or a group id option should be available as well.

lasley:
Ohhhhh that makes sense. Basically a virtual extension for the fax number, which would also implicitly provide the permissions for the inbound fax - the callback will "dial" the virtual extension, triggering the phone receipt workflow?

The only issue I see with this workflow is that faxes should be handled more akin to an email, in an asynchronous manner vs the synchronous workflow of phonecall => queue => agent. 

On my end, this will be implemented in a pharmacy workflow paired with `vertical-medical`. The customer will be receiving ~100-200 faxes per hour, which will then be dropped into an inbound queue for handling similar to the Sale Leads queue. 

Maybe it would make sense to add the user notifications as an option on the interface? Or did I just go down the entire wrong avenue of thinking?

treveradams:
No, FreeSWITCH has faxing capability built in, along with virtual modems, support for things like HylaFax. What I am seeing is you have a given pool of numbers on which faxes can be received. FreeSWITCH obviously knows the extensions or DID (destination number). Something like the notify/pop-up call back would be used to push the inbound fax into the Odoo system. At least, I believe yours does that through pull with SFax.

So, customer sends a fax to fax_number_a (FreeSWITCH here is the soft switch and I will implement it for sending and receiving over the next few months, I am just hoping to use at least some of your infrastructure). fax_number_a may be the configured fax number for the user, in which case it would call the xml-rpc for, let us call it for now, inbound_fax_by_extension. This call back would have most of the parameters your pull would have: sender number, receiver number (which may or may not be the extension depending on how FS is configured), and the PDF base64 encoded (from what I see of your code, this is what you want). Your code would then place it however yours does (when _by_extension, preferably for that user only) into Odoo, but inbound_fax_by_extension would be for a specific user.

You may want an inbound_fax_global (does what you seem to be describing) and inbound_fax_by_group_id (which is the Odoo group ID). Security should only allow the specified individual, group or global to see it. Not all work flows have all inbound faxes go to everyone, nor should they; some would have groups, some only users, some a combination of the three.

Like the popup, you would need a user that has the technical capability of receiving the faxes and storing them (instead of caller ID lookup).

I am sorry that I used the notification rpc is an example. I thought it would help clarify instead of muddy the waters. A notification could be configured, but I was just trying to explain what I had in mind. If you are willing to work with me, I will work on the FS part/odoo adapter.

lasley:
Ok got it, my bad for taking the example too literally. The intent here is definitely to allow this infrastructure to be used outside of my purposes, so I am 100% willing to work with anyone that wants in on this - the more minds the better IMO

Seems like you've got a solid grasp of what I've done here (although I ended up changing PDF storage to PNG for PIL compatibility purposes). You're also on the right track that the SFaxes are pulled in, which is triggered via the callback controller. Kind of a funky way of doing things, but I was trying to avoid cron driven at all costs.

I think I'm getting a picture of how this will work with FreeSwitch as well, thanks for the breakdown. With it being an on-premise solution, there are a lot more numbers to work with and thus the faxes will likely be more personalized. 

What about security on outbound faxes/extensions? Are the numbers pooled in FreeSwitch similar to how they are in SFax, or is it more of a 1:1 ratio?

I was also thinking of adding a generic "Fax" handler that would allow a report to be faxed directly. Was thinking of hooking into the report handler to create a PDF then fax on the fly. Would that be useful on your end as well?

lasley:
I was digging around in the import_email, and I think I'm now understanding more about why RPC. Sorry, I'm totally new to Odoo so I was confused as to why we would want to hit the external API from an internal application, but RPC is also controlling events. Something like this right https://github.com/odoo/odoo/blob/9.0/openerp/tools/import_email.py#L141 ?

treveradams:
I am relatively new as well. I started by forking the Akretion Asterisk connector to do FreeSWITCH. I haven't done much else until very recently.

I do not want to call an external API from internal. I want to call internal from an external API. The notify example is perfect in the flow, but not content.

Call comes in to FreeSWITCH or Asterisk (or goes out if you are setting callee_id_name/number), FreeSWITCH does an XML-RPC call to Odoo (OpenERP 7 in my case), such as get_caller_name.py does in my FreeSWITCH fork:

```
        if options["notify"]:
            res = sock.execute(
                options["database"], options["user"], options["password"],
                'phone.common', 'incall_notify_by_extension',
                phone_number, options["notify"])
            stdout_write('VERBOSE "Calling incall_notify_by_extension"\n')
```

user and password match the user that has the notify/caller ID lookup technical feature permission turned on. (incall_notify_by_extension is only in my code it is by_id or some such in the main tree.)

I would like to see three functions that can be called via RPC (naming is not required to match mind of course):

inbound_fax_by_extension (1:1 if only one user has this extension defined as theirs under the telephony tab, my function does allow multiple, and expects, as does my incall_notify_by_extension, that one or more users have this extension mapped as theirs) whether one to one or not the user(s) with the extension get their own private copy of the fax and any Odoo security would enforce this.

inbound_fax_by_group_id - this is 1 to many. It would work like incall_notify_by_login except for two things 1) it is group id not user id and 2) the security enforcement and delivery is to a group box.

inbound_fax_global_destination - this is a 1 to all company employees. I don't see me needing this. I don't know why anyone would. I am suggesting it for completeness. inbound_fax_by_group_id could be used to the same effect if there is a global group.

Each of these would would have several parameters:
local_fax (the fax number that received the fax)
remote_fax (where the fax is being sent to)
state (done, some other value showing the problem)
timestamp
direction would be known so it could be in by default
attempt_num (the attempt number)
page_num (I am not sure how you define this. Are you saying the number of pages sent or the number of pages that are in the fax, because on error these may be different).
status_msg (OK, what other value?)
response_num (what is this?)

As I write this I am realizing you may already have most of this done in process_in_fax/save_transmission. The problem is that I am not sure where I would get fax_id, transmission_id. I imagine these are supplied to you by SFax? If so, I would need to manufacture uuids for these values. That only leaves the three (or two) security and storage models.

As my setup would have the faxing on campus, I don't see why I would want to pull. I need FS to push (with appropriate recover if the push fails for some reason). Hence the XML or JSON RPC desirability. That and I would need to somehow accomplish fetching payloads via push instead of pull; preferably one RPC call for it all.

As for PDF or PNG, I need PDF for outgoing (think signing faxes, etc.). PNG for incoming will work, I think. The part for PDF for sending may not even need to be in anything that your code would affect. I also see yours appears to still be PDF for outgoing.

So, on second thought as I am working on my solution as I respond to yours I have the following needs:

1) Do you think you can help get this going for 7.0 as I don't have 8.0 yet? (Not required)
2) I need to understand where you are getting the following values and what they mean:
 transmission_id (each value in the structure), payload_ids, sfax_id.

Out of curiosity can you better explain what the vector, apikey (password?) are?

lasley:
Cool I think we're aligning in our thoughts here. 

It sounds like you can actually configure the callback that FreeSWITCH is sending to Odoo in order to match your specifications? I'm not so lucky with SFax, they have a very standard callback that is sent using incorrect MIME headers, thus breaks the RPC protocol :frowning: 

You're right that I've done a lot fo that work in the `process_in_fax/save_transmission`. I put it all there because I had a feeling that SFax would be very different from most other providers, which is where my funky inheritance system comes into play.

Also right about the `fax_id` - it is an internal identifier that SFax uses to represent the fax. I centralized this because I assumed that there would be some sort of unique identifier in other systems, as well. Is this not the case in FreeSWITCH?

`tranmission_id` isn't really particular to an interface, it's just the relation between `fax.payload`s (fax data, can be multiple `fax.payload.pages`) and the actual `fax.transmission`s (can be multiple per payload). I think I may have had an ambiguously named `transmission_id` field that ended up getting renamed to `fax_id` when I saw it

Hmmmm yeah on second thought `page_num` is a bit ambiguous. It is meant to represent the amount of pages that are included in the payload (could consist of cover pages, multi pages, etc). Not really anything to do with the transmissions themselves. This maybe should be removed now that I added multi page logic via the `fax.payload.page` logic.

As for `status_msg`, the only definition that SFax has provided in their API docs is `OK`. I'm still awaiting an error so I can document that better. I assume there is some sort of status that FreeSWITCH defines too though? Maybe we should come up with a general status set, then map the proprietary ones to them?

`response_num` is just the `fax_id`. Seems I forgot my naming schema there, my bad!

In the SFax callback method, I only receive the fax information as an inbound callback and am then expected to go pull the actual payload from the API from that trigger. It sounds like you're able to unify this on FreeSWITCH, but I think this should still work with the way things are structured. I'm not sure if you'd want to do this though, as it seems like you would be sending payloads unnecessarily for status updates? 

PDF vs PNG - yeah I need PDF for outbound as well, and the interface is configured as such. The problem is that PIL (python imaging library) can't read PDFs, so the internal storage format needs to be PNG in order to allow easy conversions. That's also why I had to use TIFF inbound instead of PDF (SFax only supports TIFF and PDF). Also, image formats render in views vs. PDFs don't.

Hmmm I'm not too sure what would be required for v7. I'm pretty sure I used all the backwards compatible apis, is this not the case?

Ahhh yes so the vector is for block cipher enryption/decryption. SFax provides an initialization vector (IV) and encryption key, which you can then use to authenticate and decrypt communications from them. It's an extra layer of security on top of your standard token based API authentication schema. Basically the token is encrypted with additional data like the DateTime in order to prevent extended replay attacks (eg they don't allow inbound tokens older than 15 minutes, my controller doesn't allow them after 5 minutes)

From https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation

> Most modes require a unique binary sequence, often called an initialization vector (IV), for each encryption operation. The IV has to be non-repeating and, for some modes, random as well. The initialization vector is used to ensure distinct ciphertexts are produced even when the same plaintext is encrypted multiple times independently with the same key.[6] Block ciphers have one or more block size(s), but during transformation the block size is always fixed. Block cipher modes operate on whole blocks and require that the last part of the data be padded to a full block if it is smaller than the current block size.[2] There are, however, modes that do not require padding because they effectively use a block cipher as a stream cipher.

Let me know if I missed something!

treveradams:
(I will answer peicemeal as I am able to go over what you have written.)

I wasn't sure if vector was IV or not. Thank you for clarifying.

As for FreeSWITCH, the only any ID for anything in FreeSWITCH are call and leg UUIDs. I suppose these could be fax IDs, but they are very large integers of course and are likely unique but not gauranteed.

FreeSWITCH fax returns from SpanDSP rx/txfax applications:
- '''fax_result_code''' - 0 on error otherwise >= 1;
- '''fax_result_text''' - fax error string, provide info where an error has happened;
- '''fax_success''' - 0 on error, 1 on success;

I haven't seen what the text is yet.

I was incorrect, I do need help figuring out how to do the equivalent security models to those three RPCs I suggested. I believe you are correct that it may be best to have two separate entry points the way you are doing. Is it possible to just have status_msg be a text field and store whatever the backend provides (FreeSWITCH, SFax, etc.)?

lasley:
Hmmm it doesn't sound like those leg UUIDs will be very useful in terms of identifying the fax, but there probably also needs to be a solid way to relate back to the system fax if at all possible. Does FreeSWITCH store any of the fax/metadata past the point of handing off to an API?

I agree with leaving the status_msg as a freeform field, it seems that there is insufficient documentation on all ends in order for us to do anything with unification. It would probably also be good to have the exact text just in case troubleshooting needs to take place.

Regarding the security model for RPC, there's a few ways you can likely go about it. 

Personally I would create a service account specifically to receive faxes via the RPC method. This will isolate your faxing from users that could possibly become compromised or disabled.

I would keep everything FreeSWITCH side as minimal as possible, basically just push the fax & meta into Odoo for processing. Odoo should then determine the extension, group, or global dest to direct the fax to via the metadata provided (I'm assuming the DID would probably be the best routing method here?). 

I would probably unify the three security models into two. Create a model to house the DIDs that exist. Relate to `res.users` with a Many2many in order to cover the pushing directly to user. Relate to `res.groups` also with a Many2many in order to cover the pushing globally or to groups.

Did I oversimplify?

treveradams:
I don't think you oversimplified. I think you should match the DID (whatever FreeSWITCH tells you it is) to the extension in the user AND the fax number. It may match one but not the other. I would also do it against employee accounts only. It is my understanding that partners can have accounts as well (I haven't done this, so I am not sure). I am not sure how to do it with the groups as I don't think groups have extensions/fax numbers.

Everything else seems to be correct. I will be taking a break due to the need to make a fax server so that if Odoo is down it will keep trying to notify Odoo and transfer the fax to it. Additionally, sending will require one. I don't think this is too difficult for a simpler one. I may stick a few APIs on top (Hylafax like, etc.) so that it works with as much software as possible. As this is currently a low priority for me, it may take a few months.

The account you described is what I was trying to describe. With the pop-pup/CID lookup account, it is defined by a special technical feature permission. I would recommend you do the same method as it works and doesn't have hard coded names/ID numbers.

treveradams:
Have you ported this to 9? I may be soon able to look at FusionPBX/FreeSWITCH integration.

lasley:
We have a working version for 9 [in our repo](https://github.com/laslabs/odoo-telephony).

lasley:
I'm closing this one - we're supporting v8, 9, and soon to be 10 within our [own repo](https://github.com/laslabs/odoo-telephony) due to lack of interest here. We'd obviously be willing to bring them over to OCA if interest ever peaks.

pedrobaeza:
- What is your fax number?
- I don't have fax where I live
- And where is that?
- On 21st century

:rofl: :rofl: :rofl:

lasley:
Lmao while I totally agree, it's also unfortunately the primary method of communication between doctors and pharmacies (also pharmacy to pharmacy) in America. Freaking archaic, but it's a solution that multiple companies have paid for at this point out of necessity 😑

pedrobaeza:
Yeah, it was a joke, but as you say, unfortunately not all technologies evolve the same as we like to.

PR #68, opened 2015-10-26
[FIX] Tests

StefanRijnhart:
Conversion of a phone number leads to an exception when triggered through the interface, or an error message otherwise. If the error message is not suppressed, travis will always reject the result. This does not undermine the test as the test validates that the phone number in case has not been modified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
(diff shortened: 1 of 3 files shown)


StefanRijnhart:
Additionally, fixing other problems in test by applying a safe way to query the registry for instanciated models, and test the click2dial modules in isolation so that [1] the website module's tests which check for the original phone number formatting do not interfere with base_phone's reformatting, and [2] a user form change is not loaded at a time when unrelated modules which add a group and therefore a group checkbox field are not instanciated yet.

dreispt (approved):

StefanRijnhart:
Thanks for the valiant attempt @dreispt! I do remember being a bit disappointed that this PR was neglected back in the day, but I think closing it is probably our best option after (checks date) more than 5 and a half year :laughing:

PR #69, opened 2015-10-28
Less verbose logs

guewen:
Logs are cluttered by 'initial value...updated value' and it is really not
necessary to log that using a 'info' level.
The 'error' log level is excessive for the missing country.

If someone really wants to display the original and updated values, there still
is the possibility to use a `--log-handler=openerp.addons.base_phone:DEBUG`.

### DIFF
--- a/base_phone/base_phone.py
+++ b/base_phone/base_phone.py
@@ -78,9 +78,9 @@ class PhoneCommon(models.AbstractModel):
                     countrycode = user.company_id.country_id.code
                 else:
                     _logger.warning(
-                        _("You should set a country on the company '%s' "
-                            "to allow the reformat of phone numbers")
-                        % user.company_id.name)
+                        "You should set a country on the company '%s' "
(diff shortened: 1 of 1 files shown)


alexis-via:
ok, fine for me :+1:

base_phone/base_phone.py
```diff
@@ -77,7 +77,7 @@ def _generic_reformat_phonenumbers(
                 if user.company_id.country_id:
                     countrycode = user.company_id.country_id.code
                 else:
-                    _logger.error(
+                    _logger.warning(
                         _("You should set a country on the company '%s' "
                             "to allow the reformat of phone numbers")
```
elicoidal: add period

guewen: My point was only to change the level, but ok if you want. I replaced the `%` by `,` and removed the gettext call, a log should not be translated.

lmignon:
:+1:

guewen:
thanks!

PR #55, opened 2015-08-04
8.0 sms

Viggor:
Add modules to send sms to partner with ovh configuration.
Add a module to automatic sms sending when a deliver is ready.

The original sms module come from Julius.

PR #70, opened 2015-11-02
8.0 sms update

ahilali:
make all fields on readonly when message was sent


sebastienbeau:
duplicated https://github.com/OCA/connector-telephony/pull/55

PR #74, opened 2015-11-23
Fix Nonetype error

### LINKED ISSUE #73: base_phone: error when no country is set on the company
I get this error when no country is set on the company.

```
Cannot reformat the phone number '55956595' to international format. Error message: 'NoneType' object has no attribute 'upper' 
```

guewen:
The countrycode may be None so we shouldn't call .upper() on it.

Fixes #73

### DIFF
--- a/base_phone/base_phone.py
+++ b/base_phone/base_phone.py
@@ -82,6 +82,8 @@ class PhoneCommon(models.AbstractModel):
                         "to allow the reformat of phone numbers",
                         user.company_id.name)
                     countrycode = None
+            if countrycode:
+                countrycode = countrycode.upper()
                 # with country code = None, phonenumbers.parse() will work
                 # with phonenumbers formatted in E164, but will fail with
                 # phone numbers in national format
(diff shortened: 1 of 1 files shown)


alexis-via:
For me, it doesn't really solve the pb ; instead of crashing on country_code.upper(), it will crash on parse(), because phonenumbers.parse requires a countrycode:

```
>>> import phonenumbers
>>> phonenumbers.parse('0699099347', None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/phonenumbers/phonenumberutil.py", line 2450, in parse
    "Missing or invalid default region.")
phonenumbers.phonenumberutil.NumberParseException: (0) Missing or invalid default region.
```

I just had an idea for a better solution: if the country is not set on the company (which is really bad !), we could get the lang of the user ('fr_FR'), extract the 2 last letters ('FR') and use this as countrycode. But the lang of the user may be empty... so we could still be left without countrycode. In this case, we should either:
- raise an error
- skip the parsing of phonenumber

guewen:
@alexis-via but at least it shows a correct error message, because errors in `phonenumbers` are caught:

```
Cannot reformat the phone number '5855' to international format. Error message: (0) Missing or invalid default region. 
```

Relying on the language of the user is not a good solution in my opinion. It is not unusual in Switzerland to use `fr_FR` as language instead of `fr_CH` simply because `fr_FR` is the main french language code.

alexis-via:
OK, then I'm OK :+1:

PR #77, opened 2016-01-29
8.0 FIX avoid phone numbers on 2 lines in tree view

alexis-via:
Simple idea: replace space by unbreakable space !


cdubuit:
:+1: using this PR in production

sebnege:
+1 using this patch in production, works fine

PR #84, opened 2016-04-01
8.0 avoid crash in report caused by the failure to reformat a phone number

alexis-via:
If a "phone"  field is not in E.164 format (because it was not reformatted by the wizard for example), then phonenumbers.parse(phone) inside the format() method of the report will raise an exception, which will cause a failure to generate the report.

This PR fixes this.


pedrobaeza:
Why Travis fails?

alexis-via:
It fails because it doesn't find the module web_action_request (by anybox) which is required by base_phone_popup.

pedrobaeza:
Can we have that module on OCA?

pedrobaeza:
For this PR, :+1:

PR #94, opened 2016-07-19
8.0 FIX for mac OS X

alexis-via:
Use no_break_space instead of narrow_no_break_space, to make it work on OS X.


pedrobaeza:
:+1:

MOSAMAD:
:+1: I'm using Mac OS X and I confirm that it fixes the display of phone numbers in tree views. Very good !

PR #97, opened 2016-08-04
Logcall

treveradams:
Work in progress. Please ignore for now.


treveradams:
I need to finish the lua hook for FreeSWITCH and add the configuration options for FreeSWITCH and Asterisk and fetch those. Additionally, I need to finish the partner modules and create the HR recruitment module.

treveradams:
logcall and freeswitch_logcall currently work. I believe the crm_voip_logcall and asterisk_logcall work as well, but I don't have the ability to test. I am still working on the rest of the modules. Of course, these depends on my FreeSWITCH Odoo 9.0 branch.

treveradams:
The FreeSWITCH scripts in the logcall directory are currently broken. The python one will be removed as it is just a way for me to keep track of things. The lua one may work, but hasn't been finished or tested. That one is what will remain.

treveradams:
The work remaining at this point is cleanup, making sure asterisk and FreeSWITCH provide the proper dates. Finishing the lua hook for FreeSWITCH.

treveradams:
This is being used locally. It only has two bugs.

1) Calls for hr_recruitment have a contact that is invalid. This is because of the use of partner_id to store id of applicant. I have not yet found a way around this.

2) If using FreeSWITCH and answering party is on a FreeTDM device, you get the span number instead of internal extension. This may be a problem with FS or with my installation.

I would greatly appreciate testing by Asterisk users and FreeSWITCH users.

treveradams:
Much of this work was done by others. https://github.com/credativUK/connector-telephony/commit/900f9d310e634958801d337987193dc64845ae45#commitcomment-18471261 gave me permission to submit it as my own provided I gave credit.

treveradams:
The hr_recruitment/hr_applicant bug has been quashed. The other is an issue either with FS or with my setup, not with this code.

Again, any Asterisk or FS users, please test. FS users, it works best if all of your resources (id's in the directory) are numbers and not alpha-numeric. Be consistent either way.

treveradams:
I have drastically cleaned up the code. I have added a few features. I believe this is now feature complete as far as the Odoo code. I am still working out a few kinks with the freeswitch_logcall.lua script. I am still encouraging testing. Documentation and security models have been added/updated.

treveradams:
The only problem I can find with this module at this point is if it is called two times with the same data so fast that one request to the database starts before the other finishes. This only prevents the calls from being merged as one (duplicate exists), not the recording of the data.

The other has to do with if the call coming in is bridged (not transferred) to a FreeTDM channel. The destination number in that case will be the channel number not the phone number assigned. This is a problem with FreeSWITCH that I cannot overcome. I do not have an Asterisk setup, so I cannot test that.

PR #103, opened 2016-09-05
[FIX] crm.phonecall. default in team_id should be a browse 'crm.team', not an int

JordiBForgeFlow:
Otherwise error occurs when loading the module 'crm.phonecall' with message "AttributeError: 'int' object has no attribute 'id'"


alexis-via:
:+1: thanks for the fix.

AaronHForgeFlow (approved):
Don't know why the runbot is failing. Tested in local + code review :+1:

ecino:
The fix makes migration to v9 work, failing without it. Should be merged in my opinion. Thanks for that 👍

ecino (approved):

MiquelRForgeFlow (approved):
LGTM :+1:

PR #105, opened 2016-10-02
8.0 Import phonenumbers protected by try/except

alexis-via:
This PR is designed to avoid Travis errors such as this one:
https://travis-ci.org/OCA/sale-workflow/jobs/164489455
The sale-workflow project doesn't include connector-telephony via oca_dependencies.txt, but it includes account-invoicing which has "connector-telephony" in oca_dependencies.txt. So the Travis of sale-workflow has connector-telephony in its addons_path but it doesn't have the phonenumbers lib... which triggers the problem.

### DIFF
--- a/base_phone/base_phone.py
+++ b/base_phone/base_phone.py
@@ -24,7 +24,10 @@ from openerp.tools.safe_eval import safe_eval
 from openerp.exceptions import Warning
 import logging
 # Lib for phone number reformating -> pip install phonenumbers
-import phonenumbers
+try:
+    import phonenumbers
+except ImportError:
+    logger.warning('Cannot import phonenumbers')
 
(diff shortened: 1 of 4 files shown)


bealdav:
:+1:

bguillot:
:+1:

base_phone/base_phone.py
```diff
@@ -24,7 +24,10 @@
 from openerp.exceptions import Warning
 import logging
 # Lib for phone number reformating -> pip install phonenumbers
-import phonenumbers
+try:
+    import phonenumbers
+except ImportError:
+    logger.debug('Cannot import phonenumbers')
```
sebastienbeau: logger is not define ;)

PR #107, opened 2016-10-08
[MIG] Migrated base_phone to v10.0

Garamotte:
Some notes about the migration :
- The phone and fax widgets are now extending the email widget in Javascript, to beneft from the prefix attribute.
- In XML, the widgets' templates extend FieldEmail using t-call instead of t-extend because the later doesn't seem to work (maybe a standard Odoo bug).
- Click2dial part still untested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,6 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
   - pip install phonenumbers py-Asterisk SOAPpy
-  - hg clone http://bitbucket.org/anybox/web_action_request -b ${VERSION} ${HOME}/web_action_request
-  - if [ -d /home/travis/${ODOO_REPO##*/}-${VERSION}/addons ]; then ln -s ${HOME}/web_action_request/web_action_request /home/travis/${ODOO_REPO##*/}-${VERSION}/addons; fi
-  - hg clone http://bitbucket.org/anybox/bus_enhanced -b ${VERSION} ${HOME}/bus_enhanced
-  - if [ -d /home/travis/${ODOO_REPO##*/}-${VERSION}/addons ]; then ln -s ${HOME}/bus_enhanced/bus_enhanced /home/travis/${ODOO_REPO##*/}-${VERSION}/addons; fi
 
 script:
   - travis_run_tests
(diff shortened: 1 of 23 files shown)


alexis-via:
Great work. We still have to fix one think: in the URL "tel:+33141984242", the spaces/non-breaking-spaces are still there ; they should be removed. Once it is fixed, it can be merged: :+1:

Garamotte:
Number format in the link fixed.
Shouldn't it be the "raw" format ? And adding spaces/dashes/other would be done only for display.

I also added a new commit, to define Phone and Fax as real field types.
This avoids the need of redefining the "phone" or "fax" widget in each view.
I kept this commit separated because I'm not sure it won't create side effects.

alexis-via:
@sylvain-garancher It's great that you found a solution to avoid inheriting the views to add widget="phone" or widget="fax"... something I wanted to do for a long time (or at least since v9).

I tried your latest code, but when I try to access the form view of a partner, I get a JS error
"TypeError: phone_num.replace is not a function"

Garamotte:
Oops, sorry, I forgot to test with empty values :)

I added another change : The phone field now inherits from fax, as it has the same behaviour, plus the clcik2dial feature. This avoids duplicated code between the two field types and widgets.

alexis-via:
Works very well now ! :+1:

.travis.yml
```diff
@@ -32,10 +32,10 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
   - pip install phonenumbers py-Asterisk SOAPpy
-  - hg clone http://bitbucket.org/anybox/web_action_request -b ${VERSION} ${HOME}/web_action_request
-  - if [ -d /home/travis/${ODOO_REPO##*/}-${VERSION}/addons ]; then ln -s ${HOME}/web_action_request/web_action_request /home/travis/${ODOO_REPO##*/}-${VERSION}/addons; fi
-  - hg clone http://bitbucket.org/anybox/bus_enhanced -b ${VERSION} ${HOME}/bus_enhanced
-  - if [ -d /home/travis/${ODOO_REPO##*/}-${VERSION}/addons ]; then ln -s ${HOME}/bus_enhanced/bus_enhanced /home/travis/${ODOO_REPO##*/}-${VERSION}/addons; fi
+  # - hg clone http://bitbucket.org/anybox/web_action_request -b ${VERSION} ${HOME}/web_action_request
```
pedrobaeza: We should directly remove these lines.

pedrobaeza (approved):
Tested in runbot. Little thing for merging

Garamotte:
Commented lines removed in TavisCI configuration.

PR #109, opened 2016-10-28
[10.0] Port most modules and fix bug 117

alexis-via:
Port crm_phone, hr_phone, hr_recruitment_phone, event_phone to v10
Remove crm_claim_phone, because there is no crm_claim module in v10
Start to port asterisk_click2dial: JS code still needs to be ported. @sylvain-garancher I'll probably need your help on this ; you have much more experience than me in JS and a big part of the work is to adapt the JS code of asterisk_click2dial to the changes you made in the JS code of base_phone.


pedrobaeza:
Please don't remove _crm_claim_phone_, as we plan to maintain crm_claim module as an OCA one (as we did with _contract_), because there's no other alternative for now. Indeed, helpdesk approach is better, but the one provided by Odoo is enterprise and there's no other replacement for now, so the quickest migration path from our customers that are in v8/v9 is to keep it.

alexis-via:
OK, no pb. But is crm_claim already in the OCA or not yet ?

pedrobaeza:
Not yet. We will port it when we start to move customers in 8.0/9.0 to 10.0 (6 months or so).

alexis-via:
click2dial is working now.
Next step : fix the display "open caller" button (phone logo) at the top right ... and make the corresponding feature work.

alexis-via:
For me, with my latest commit, everything is working now (except the bug #114  reported by invitu). Any feedback/test is welcomed.

alexis-via:
Runbot is red only because it misses the py-Asterisk lib... don't we have a way to define the python libs required by both Travis and runbot ? I think we have something for that, but I can't find an example.

Garamotte:
You can add a requirements.txt file on the repo's root.
See in server-tools.

alexis-via:
Mmm, it didn't work, it's all red now, even Travis which was green before the change.

alexis-via:
OK, Travis was red for some other reasons. I also found out that there was already some "requirements.txt" files inside the base_phone and asterisk_click2dial directories... but I don't know if it works if it's inside a module directory.

alexis-via:
@gurneyalex If you have a clue why runbot didn't install the "py-Asterisk" lib although I added a file "requirements.txt" at the root of the repo, please tell us. I'm trying to make runbot green, but it fails just for this missing lib.

pedrobaeza:
That's because this repository is not yet using Travis2docker. I have changed the configuration in runbot and launch the rebuild to see.

You might also improve coverage on the repository.

Garamotte:
@pedrobaeza The runbot build was skipped.
Do you know why ?

pedrobaeza:
Uhm, that seems a general issue that they're still handling. See https://github.com/OCA/runbot-addons/issues/105

gurneyalex:
@alexis-via should be solved now.

alexis-via:
I received some positive feedback of one of my users who uses the asterisk connector in production on odoo10... Travis is green. So let's merge :)

diggy128:
This module breaks phonecall as it redefines the crm.phonecall model that is declared in crm_phonecall in crm repository.

PR #110, opened 2016-10-29
[FIX] Fix point_of_sale partner error

cvinh:
When updating a partner in the frontend of the POS Odoo generate a write() with the ID of the country as unicode !!!
example : vals = {u'country_id': u'9'}
So we have to convert it to an integer before browsing

### DIFF
--- a/base_phone/fields.py
+++ b/base_phone/fields.py
@@ -125,6 +125,12 @@ def write(self, vals):
     fields_to_convert = get_phone_fields(self, vals)
     if fields_to_convert:
         for record in self:
+            # When updating a partner in the frontend of the POS
+            # Odoo generate a write() with the ID of the country as unicode !!!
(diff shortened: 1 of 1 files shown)


alexis-via:
I confirm the bug and I confirm that the PR fixes the issue :+1:

alexis-via:
I propose an alternative patch in this commit: https://github.com/akretion/connector-telephony/commit/e590672c30b7fe8cc49394642fc1fd415fbda351
I think my solution is lighter because it avoids a try/except, which doesn't seem necessary to me.

This commit is included in my PR https://github.com/OCA/connector-telephony/pull/109

cvinh:
thanks @alexis-via 
I close this PR
Could you backport it for v9 ? my #111  does not pass the travis tests

Issue #114, opened 2016-11-08
[10.0-base_phone] Including base_phone in addons gives error in creating project


cvinh:
Steps to reproduce : 
- add base_phone in addons
- install project_project module
- create a new project
- error : 
```
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: alias_model_id - alias.model.id]
```
The problem seems to be in base_phone/fields.py (method create())

alexis-via:
I confirm the bug on v10. v9 is not affected. The only known workaround so far is to uninstall base_phone AND remove it from the addons_path (uninstall is not enough, because it use monkeypatching).

The pb comes from the fact that, in addons/mail/models/mail_alias.py (https://github.com/odoo/odoo/blob/10.0/addons/mail/models/mail_alias.py#L101), the create() method is inherited and it uses 

```
model_name = self._context.get('alias_model_name')
[...]
if model_name:
     model = self.env['ir.model'].search([('model', '=', model_name)])
      vals['alias_model_id'] = model.id
```

When base_phone is in the addons_path, the context doesn't contain the 'alias_model_name' key anymore, so alias_model_id is not added to vals.

@sebastienbeau @sylvain-garancher If you have an idea about the root cause of this problem, please make some suggestions here !

pedrobaeza:
Can we monkey-patch in the _register_hook method or inheriting base model like in https://github.com/OCA/server-tools/blob/10.0/base_suspend_security/models/base.py so that we don't need to remove the folder?

alexis-via:
@pedrobaeza I'm not sure to understand you last comment ; it is a solution to fix the bug or just a suggestion to avoid the need to remove connector-telephony from addons_path ?

Garamotte:
I just created a new pull request that fixes this bug.

Since v10, we must inherit from the base abstract model instead of monkey-patching methods from models.Model.

cvinh:
I confirm the fix works for us

PR #115, opened 2016-11-08
[FIX] We must now inherit from the base model, instead of monkey-patching

Garamotte:
Fixes https://github.com/OCA/connector-telephony/issues/114

cc @alexis-via @invitu


pedrobaeza (approved):

cvinh:
Hi @sylvain-garancher your fix works
Thanks

alexis-via (approved):
Successfully tested : I confirm it fixes the bug.

Issue #113, opened 2016-10-30
[base_phone] Problem in point_of_sale with editing partners


cvinh:
PR for v10.0 here : https://github.com/OCA/connector-telephony/pull/110

cvinh:
PR for 9.0 here : https://github.com/OCA/connector-telephony/pull/111

cvinh:
fixed for v10 in #109

alexis-via:
I added a unitest in my PR #109 to make sure that it doesn't happen any more.

PR #118, opened 2016-12-19
[8.0] Restore fix for partner creation from POS bug #113

StefanRijnhart:
Backport of https://github.com/OCA/connector-telephony/pull/109/commits/6c68f3e7cf86a2b0

### DIFF
--- a/base_phone/base_phone.py
+++ b/base_phone/base_phone.py
@@ -56,8 +56,11 @@ class PhoneCommon(models.AbstractModel):
             countrycode = None
             if self._country_field:
                 if vals.get(self._country_field):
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Uhm, Travis fails...

Maybe we should protect if the value is False?

StefanRijnhart:
I guess we can simply backport all the work on 9.0/10 for this module to properly fix travis, but that would be out of scope for this change.

We don't need to protect against falsy values, because both calls are preceded by a condition 'vals.get()'.

PR #149, opened 2017-10-05
[FIX] base.phone : add default company on sms model

fmdl:
There are some rules, but if the company field is not set, nobody can see SMS : https://github.com/OCA/connector-telephony/blob/10.0/base_sms_client/security/ir.rule.csv


fmdl:
@alexis-via

PR #146, opened 2017-09-25
[ADD] odoo 10 portage of base_phone_validation

ventilooo:
A v10 migration of #42

PR #162, opened 2018-03-23
[11.0] base_phone_validation: version 11 of the module

foutoucour:
Work to port the module to version 11
Seems the work done on #146 did most of the job.


pedrobaeza:
Odoo core already includes this feature. Is this still needed?

foutoucour:
True. It is not needed it seems.

PR #167, opened 2018-09-21
[MIG] base_phone_popup: Migration to 11.0

cubells:
- Remove external modules dependencies and use standard phone_validation modules.
- keep `number_of_digits_to_match_from_end` parameter in company to search the last n digits on phones.
- hook method to define in which fields of models to search

Must be tested

cc @Tecnativa


base_phone_popup/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2014-2016 Akretion, Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Base Phone Pop-up',
+    'version': '11.0.1.0.0',
+    'category': 'Phone',
+    'license': 'AGPL-3',
+    'summary': 'Pop-up the related form view to the user on incoming calls',
+    'author': "Akretion,Odoo Community Association (OCA)",
```
pedrobaeza, line 10: Tecnativa

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class PhoneCommon(models.AbstractModel):
+    _inherit = 'phone.validation.mixin'
+
+    @api.model
+    def get_record_from_phone_number(self, presented_number):
+        """If it finds something, it returns (object name, ID, record name)
+        For example : ('res.partner', 42, u'Alexis de Lattre (Akretion)')
```
pedrobaeza, line 17: Why don't simply return the record?

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        _logger.debug(
+            "Call get_name_from_phone_number with number = %s"
+            % presented_number)
+        if not presented_number.isdigit():
+            _logger.warning(
+                "Number '%s' should only contain digits." % presented_number)
+        nr_digits_to_match_from_end = \
```
pedrobaeza, line 25: I'm not sure if this should be at company level, or system wide (through a system parameter). What can be the reasoning for having different lengths for company?

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+            self.env.user.company_id.number_of_digits_to_match_from_end
+        if len(presented_number) >= nr_digits_to_match_from_end:
+            end_number_to_match = presented_number[
```
pedrobaeza, line 28: I think the logic should be the contrary: you want to strip extra numbers at the end, not at the beginning. Any way, as it is, you don't need to make the `if`: `presented_number[-nr_digits_to_match_from_end:]` always return correct thing. But as said, I doubt this is the best approach. @alexis-via what is the reason for stripping initial numbers?

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+                return res
+            else:
+                _logger.debug(
+                    "No match on %s for end of phone number '%s'",
+                    obj._name, end_number_to_match)
+        return False
+
+    @api.model
+    def _get_phone_models(self):
+        # Hook to get models and fields on which search
```
pedrobaeza, line 67: Put this as docstring, not as comments, and specify the used format.

base_phone_popup/models/phone_common.py
pedrobaeza: Reading the field help, I would say that the technique should be to normalize the inbound number before comparing. What I said about stripping final numbers is because some switchboard add the extension number at the end, and this is something that we can handle searching first for full match, and if not, try to strip that part.

base_phone_popup/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,6 @@
+* Akretion <https://akretion.com>
+
+  * Alexis de Lattre <alexis.delattre@akretion.com>
+* Tecnativa <https://www.tecnativa.com>
```
pedrobaeza, line 4: Put URL format

pedrobaeza (changes requested):
Where this popup is called? I understand that on subsequent modules, isn't it?

base_phone_popup/models/phone_common.py
alexis-via: The reason to strip initial number is the following: from my experience, calls from abroad are regularly badly presented by some telco without the country prefix, etc... So matching on the end of a phone number avoid this kind of trouble on incoming calls with foreign numbers without proper country prefix.

pedrobaeza: OK, we can put both options.

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+                'view_mode': 'form,tree',
+                'views': [[False, 'form']],  # Beurk, but needed
+                'target': 'new',
+                'res_id': record_res[1],
+                }
+        else:
+            action = {
+                'name': _('Number Not Found'),
+                'type': 'ir.actions.act_window',
+                'res_model': 'number.not.found',
```
alexis-via, line 92: You're calling the "number.not.found" wizard defined in base_phone but you don't depend on base_phone. BTW, it's one of the things base_phone is useful for :)

pedrobaeza: OK, I see. When are you going to publish `base_phone` to v12? So that we can backport it and use it as dependency for this (we will later return the favor migrating this to v12).

alexis-via: I ported the main phone module to v12 (base_phone, asterisk_click2dial, crm_phone, hr_phone, hr_recruitment_phone, event_phone), starting from the v10 code. As I couldn't make a PR because the current v12 branch had nothing in common with my v12 branch coming from v10, I pushed it directly, and now we can improve/make PR, etc...

This morning, I made a quick backport of that work from v12 to v11. It's starting to work well; I should be able to finish the tests/cleanup in the coming days.
@pedrobaeza If you are OK with that, I will push my backport directly to origin/11.0 like I did for v12, and then we can make PRs on it. It may not be a standard OCA way of doing such things, but if we want to move forward on v11 for connector-telephony, I think it's the right way to move forward and unlock this project.

pedrobaeza: @alexis-via you should do a PR, although you merge yourself when CIs are green and several days have passed, but at least do that for:

* not polluting main branch with several trials for greenifying CIs.
* let the opportunity to comment some things.

One thing you are not going to attend for sure is the commit squashing for reducing the useless ones, so let also that others like me do that task.

And please one PR per module.

alexis-via: I won't do 1 PR per module, because they all depend one on each other, it's too much (useless) work. These modules work together, they must be tested together, so it's much better to put them in the same PR.

How do you make a PR if there is no common history? I tried to make a PR yesterday night for v12, but I couldn't find a way, that's why I decided to push directly.

pedrobaeza: @alexis-via they don't depend each other. It's a hierarchy. Look how we are migrating OCA/bank-payment. About the PR, you just need to follow the procedure in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0, changing remotes if you are going to migrate from v12.

alexis-via: I don't have time for that for this project.
Nobody made the hard work required to port these modules to v11, and we are already 13 months after the release of v11 (if I exclude the work of one contributor who didn't keep commit history, so it couldn't be accepted). I made that work. Now, we need to move forward and not loose our time doing one PR per module. It's too much work, it takes too much time. Maybe it can be justified for large projects such as OCA/bank-payment, but not for small projects with low level of contributors like OCA/connector-telephony. The 6 modules I ported work together and need to be tested together. Code review is good, but real-life testing is also very important, and for that one PR with all the modules is much easier.

pedrobaeza: @alexis-via if you need to try all together is because they are not enough decoupled or good unit testing. Any way, you are going to do what you want as always...

alexis-via: Unit testing is difficult because you need an asterisk server to make real tests!

pedrobaeza: Well, the usual technique is to mock that things. Don't misunderstand me. I thank a lot your efforts to migrate this to v11, but I think you always make the same kind of block merging (like OCA/bank-payment in v10), and I'm trying to prove you that it can be done by parts. On your PRs, we can only perform a leap of faith and merge it as is, and later make consecutive PRs for solving certain things. I let you to decide and we'll do the best reviewing the work.

base_phone_popup/models/phone_common.py
```diff
@@ -0,0 +1,122 @@
...
+            [('login', 'in', login_list)])
+        _logger.debug(
+            'Notify incoming call from number %s to users %s'
+            % (number, users.ids))
+        action = self._prepare_incall_pop_action(res, number)
+        if action:
+            for user in users:
+                if user.context_incall_popup:
+                    self.sudo(user.id).env['bus.bus'].sendone(
+                        '%s_%d' % (self._name, self.env.uid), action)
```
alexis-via, line 114: @cubells  Did you test this PR in real life ? I'm not an expert of bus.bus, but I don't see how the above line of code could work. It seems you got inspiration from Anybox's module web_action_request on which base_phone_popup used to depend, and you took the code of the notify() method of that module, but you missed that this module has JS code that you didn't get here (and other small differences)...
On my side, I didn't succeed in making the above line of code work.

sbidoul: I have web_notify in oca for that kind of things.

pedrobaeza: Uhm, weird, this PR is supposed to use that module... (at least that were the requirements in the task). I haven't checked the PR deeply. @cubells can you clarify?

alexis-via: I ported base_phone_popup to v10 using web_notify and I am very pleased with the result, cf this PR https://github.com/OCA/connector-telephony/pull/171/files  I think it's much better than the previous behavior which was too intrusive.
I think we should do the same for v11/v12. But, before that, someone needs to up-port the "interactive notification" feature of web_notify on which we rely (feature added by @guewen a few months ago, cf https://github.com/OCA/web/pull/970).

PR #171, opened 2018-11-18
[MIG] base_phone_popup to v10

alexis-via:
We now use the great OCA module web_notify (OCA/web)


base_phone_popup/models/phone_common.py
```diff
@@ -44,26 +43,21 @@ def incall_notify_by_login(self, number, login_list):
...
-                if user.context_incall_popup:
-                    self.sudo(user.id).env['action.request'].notify(action)
-                    logger.debug(
-                        'This action has been sent to user ID %d: %s'
-                        % (user.id, action))
+            title = _('Incoming call')
+            message = _('Call from %s') % number
+            action_link_name = res and res[2] or action['name']
+            users.notify_info(
+                message, title, True, False, action, action_link_name)
```
guewen: Would be more readable to use the keyword arguments (`title=title, sticky=True, action=action, action_link_name=action_link_name`)

So the reader don't have to read the implementation to know what True and False refers to, also then you don't have to pass `show_reload` (default value)

base_phone_popup/__manifest__.py
```diff
@@ -1,43 +1,25 @@
...
-
-Warning : proxying WebSockets is only supported since Nginx 1.3.13 ; the
-feature provided by this module won't work with older versions of Nginx.
-
-TODO : document this new feature on the Akretion Web site :
-http://www.akretion.com/products-and-services/openerp-asterisk-voip-connector
+When the user receives a phone call, Odoo will display a notification
+at the top-right of the screen that contains a link to the corresponding
+partner/lead/employee/... and a link to the *Number not found* wizard.
 """,
```
guewen: OCA readme would be nice

guewen (approved):
Nice, a couple of nitpicks in the comments

PR #182, opened 2019-04-06
[11.0][add][clicksend_sms_client]

JordiBForgeFlow:
This module allows to send SMS messages using the ClickSend Gateway.

This PR also implements the base_sms_client for v11.

### DIFF
--- a/base_sms_client/wizard/mass_sms.py
+++ b/base_sms_client/wizard/mass_sms.py
@@ -18,6 +18,11 @@ class WizardMassSms(models.TransientModel):
     def _default_get_partner(self):
         if self._context.get('active_model') == 'res.partner':
             return self._context.get('active_ids')
+        else:
+            rec = self.env[self._context.get('active_model')].browse(
+                self._context.get('active_id'))
+            if hasattr(rec, 'message_follower_ids'):
+                rec.message_follower_ids.mapped('partner_id')
 
     gateway_id = fields.Many2one(
         'sms.gateway',
@@ -76,7 +81,27 @@ class WizardMassSms(models.TransientModel):
     @api.multi
     def send(self):
         sms_obj = self.env['sms.sms']
-        partner_obj = self.env['res.partner']
-        for partner in partner_obj.browse(self._context.get('active_ids')):
+        for partner in self.partner_ids:
             vals = self._prepare_sms_vals(partner)
             sms_obj.create(vals)
+        src_model = self.env[self._context.get('active_model')]
+        if hasattr(src_model, 'message_follower_ids'):
+            rec = src_model.browse(self._context.get('active_id'))
+            rec.message_post(body=self.message)
+            self.env['bus.bus'].sendone(
+                'refresh', [self.env.cr.dbname, self._name, self._uid])
+        return {'type': 'ir.actions.act_window_close'}
+
+    def redirect_to_sms_wizard(self, **kwargs):
+        id = kwargs.get("id")
+        model = kwargs.get("model")
+        action = self.env['wizard.mass.sms'].action_wizard_mass_sms()
+        action['src_model'] = model
+        action['domain'] = [('res_id', '=', id)]
+        return action
+
+    @api.model
+    def action_wizard_mass_sms(self):
+        action = self.env.ref(
+            'base_sms_client.action_wizard_mass_sms').read()[0]
+        return action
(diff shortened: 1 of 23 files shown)


pedrobaeza:
What about `sms` standard module?

JordiBForgeFlow:
@pedrobaeza the 'sms' relies on iap module (in-app purchases) for authentication.

This module relies on the 'keychain' OCA module to store credentials.

I will try to reuse the wizard provided by the sms module, which is almost the same as in this module

JordiBForgeFlow:
We have improved the usability and now you can send an SMS from the chatter.

image


image

The SMS message will be logged as a note.

base_sms_client/wizard/mass_sms.py
```diff
@@ -18,6 +18,11 @@ def _default_get_gateway(self):
     def _default_get_partner(self):
         if self._context.get('active_model') == 'res.partner':
             return self._context.get('active_ids')
+        else:
+            rec = self.env[self._context.get('active_model')].browse(
+                self._context.get('active_id'))
+            if hasattr(rec, 'message_follower_ids'):
+                rec.message_follower_ids.mapped('partner_id')
```
daramousk, line 25: 1) I think you might need to ` return rec.message_follower_ids.mapped('partner_id').ids` here. 
2) If you want to check whether a model extends 'mail.thread', you can also use the more succinct https://github.com/OCA/OCB/blob/11.0/odoo/modules/registry.py#L215 and see if your model is there.

base_sms_client/wizard/mass_sms.py
```diff
@@ -76,7 +81,32 @@ def _prepare_sms_vals(self, partner):
     @api.multi
     def send(self):
         sms_obj = self.env['sms.sms']
-        partner_obj = self.env['res.partner']
-        for partner in partner_obj.browse(self._context.get('active_ids')):
+        for partner in self.partner_ids:
             vals = self._prepare_sms_vals(partner)
             sms_obj.create(vals)
+            src_model = self.env[self._context.get('active_model')]
```
daramousk, line 87: https://github.com/OCA/OCB/blob/11.0/odoo/models.py#L4377
Accessing the context like that might work, but looking at `models.py` it seems that this is obsolete maybe use `self.env.context.get` instead?

base_sms_client/wizard/mass_sms.py
```diff
@@ -76,7 +81,32 @@ def _prepare_sms_vals(self, partner):
...
+            if hasattr(src_model, 'message_follower_ids'):
+                rec = src_model.browse(self._context.get('active_id'))
+                self.post_message(rec)
+        return {'type': 'ir.actions.act_window_close'}
+
+    def post_message(self, rec):
+        channel = 'wizard.mass.sms'
+        message = _("SMS Message (%s) sent to %s" % (
+            self.message, ', '.join(self.partner_ids.mapped('name'))))
```
daramousk, line 96: Does translation here make sense though? From what I understand, maybe translating only the message would suffice. Because judging by this https://github.com/OCA/OCB/blob/11.0/odoo/tools/translate.py#L442 you are going to need a lot of translated string, for every combination of partner(s) and message.

clicksend_sms_client/models/sms_gateway.py
```diff
@@ -0,0 +1,77 @@
...
+            #    retrieve = keychain.suspend_security().retrieve
+            # else:
+            #    retrieve = keychain.retrieve
+            retrieve = keychain.retrieve
+            accounts = retrieve(
+                [['namespace', '=', CLICKSEND_KEYCHAIN_NAMESPACE]])
+            return accounts[0]
+
+
+class SmsSms(models.Model):
```
daramousk, line 36: This should probably be defined in its own file.

clicksend_sms_client/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+To configure this module, you need to:
+
+* go to settings > technical > telephony > gateway configuration and select ClickSend http in method
```
daramousk, line 3: If the users are going to use a specific method of a specific api, wouldn't it be easier for them to have the `self.gateway_id.url` by default set to the api's url? And then they will have the opportunity to change it of course. Also, I think it would save them some trouble down the road if, on the click of `Save` on the create/write of a `sms.gateway` record, for a verification to be shown to the user of whether the data they provided are correct? By invoking the `Account Balance` functionality for example.

clicksend_sms_client/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Valentin Chemiere
```
daramousk, line 1: Missing Valentin's email here, not sure if that is intentional.

Issue #186, opened 2019-05-13
connector_voicent


max3903:
This module allows you to connect Odoo with Voicent (https://www.voicent.com) and is meant to be extended to integrate Odoo records and processes with phone calls made by Voicent.

# Usage

* Go to Connectors > Backends > Voicent Backends
* Create a new Voicent Backend with the host and port
* Create Call Lines to determine when (which stage in the process) calls are added to the queue
* Create Time Line to determine when (what time) calls are made

# Technical Specifications
- [x] Add a boolean field "Accepts Calls" (can_call) on res.partner
- [x] Create a Voicent Backend Call Line object (backend.voicent.call.line) with:
  - [x] a char field "Name"
  - [x] a selection field "Applies on" with an empty list
  - [x] a char field "Voicent App"
  - [x] a many2one field "Backend" (backend_id) to backend.voicent
- [x] Create a Voicent Backend Time Line object (backend.voicent.time.line) with:
  - [x] a char field "Name"
  - [x] a time field "Time"
  - [x] a many2one field "Backend" (backend_id) to backend.voicent
- [x] Create a Voicent Backend object (backend.voicent) with:
  - [x] a char field "Host" (host)
  - [x] an integer field "Port" (port)
  - [x] a one2many field "Call Lines" (call_line_ids) to backend.voicent.call.line
  - [x] a one2many field "Call Times" (time_line_ids) to backend.voicent.time.line
  - [x] a datetime field "Next Call" (next_call)
- [ ] Add a "Waiting Voicent Status" to the job's list of status
- [ ] Add a scheduled action to update the Next Call of the backend based on the time of today and the time_line_ids.
- [ ] Add a scheduled action to check the Voicent status of the running jobs
  - [ ] Get all the jobs in the status "Waiting Voicent Status"
  - [ ] Call Voicent API to check the status of the campaign of the job using information from the backend and the job

PR #191, opened 2019-05-30
[ADD] connector_voicent

murtuzasaleh:
#186 

Voicent Connector
==============
This module allows you to connect Odoo with Voicent <https://www.voicent.com> and is meant to be extended to integrate Odoo records and processes with phone calls made by Voicent.


murtuzasaleh:
@max3903 Can you please review this PR.

connector_voicent/models/backend_voicent.py
```diff
@@ -0,0 +1,112 @@
+# Copyright (C) 2019 Open Source Integrators
+# <https://www.opensourceintegrators.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+import datetime
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
+from pytz import timezone
+from odoo.addons.connector_voicent.examples import voicent
```
max3903:
```suggestion
from .examples import voicent
```

connector_voicent/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Maxime Chambreuil <mchambreuil@opensourceintegrators.com>
+* Youness MAAFI <youness.maafi@gmail.com>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
```
max3903:
```suggestion
* Murtuza Saleh <murtuzasaleh@serpentcs.com>
```

oca_dependencies.txt
```diff
@@ -1 +1,4 @@
-server-tools
\ No newline at end of file
+server-tools
+connector
+queue
+
```
max3903:
```suggestion
```

Please sort the file alphabetically.

max3903:
@murtuzasaleh Looks good. I will run some tests next week.

murtuzasaleh:
@max3903  


1) I don't think we need this condition because it is worked based on next call so when the job is created pass the Next Call value in `eta` so the job is to execute based on that.



2) When the new job is created how we can pass the date in `eta` field for the execution of this code because we can not pass Next Call field value on it, and job is execution part does not depend on the Scheduler, we can delay it based on `eta` and pass it where the method is defined.

Correct me if I am wrong.

max3903:
@murtuzasaleh 

1. Correct.
2. The job is created with the link to the backend (where the next call value is stored) so the job knows at what time it should be executed.

max3903 (approved):

patrickrwilson (approved):

PR #196, opened 2019-07-10
[FIX] asterisk_click2dial: Travis errors

max3903:


rvalyi (approved):

asterisk_click2dial/scripts/set_name_agi.py
```diff
@@ -206,7 +206,7 @@ def geolocate_phone_number(number, my_country_code, lang):
 def convert_to_ascii(my_unicode):
     '''Convert to ascii, with clever management of accents (é -> e, è -> e)'''
     import unicodedata
-    if isinstance(my_unicode, unicode):
+    if isinstance(my_unicode, bytes):
```
pedrobaeza: Are you sure this works for `bytes` or this should be decoded?

alexis-via: This raises the question of the script set_name_agi.py: should this script be run with python2 or python3 ? Reminder: this script is designed to be executed on the asterisk server, not on Odoo. So far, this script is still for python2; if we decide to convert it to python3, there may be other changes and we have to test it.

alexis-via: To avoid issues with Travis, as this script is not for Odoo, we could exclude it from travis checks.

pedrobaeza: You can put a directive at the beginning of the file for that, but I don't remember just now the syntax

PR #197, opened 2019-07-10
[12] base_phone: Improve layout of General Settings page

alexis-via:


base_phone/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     'name': 'Base Phone',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
pedrobaeza, line 8: Now with the OCA Bot Merger, there's no need to manually bump version. Even more, in this case it would be the minor version to change.

pedrobaeza:
Can you put a before and after screenshot?

alexis-via:
After the change:

base_phone_after

I thought it was broken before the change, but it seems I didn't have the last version of base_phone on the DB where I saw it as broken. So it's more an "improvement" than a fix.

pedrobaeza (approved):

pedrobaeza:
Travis fix in #196 please review

PR #200, opened 2019-07-22
11.0 Onchange Methods in ResPartner to format phone, and mobile numbers

oscars8a:
Good Morning,
I'm Oscar from SDi Soluciones (https://sdi.es).

I think it is necessary to include these onchange methods, because otherwise, and from what I have verified, the change to the international format is not made, adding the country prefix and extensions.

I have not found another way to include the extensions with the Odoo base modules.

Reviewing the code in version 12.0, I have seen that the onchange were included.

A greeting.

### DIFF
--- a/base_phone/models/res_partner.py
+++ b/base_phone/models/res_partner.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 
-from odoo import models
+from odoo import models, api
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Hello, Óscar,

Good to see you around. Maybe @alexis-via can tell about this PR. In principle, it seems OK, but I don't know deep enough this module to say.

alexis-via:
This is something that was discussed during the last code sprint in June in Spain : in the official addons of Odoo SA, the module "phone_validation" doesn't provide the onchange ; they are provided by "crm_phone_validation"... but this module depend on "crm". Please see this comment, which explains the solution we choose: https://github.com/OCA/connector-telephony/blob/12.0/base_phone/models/res_partner.py#L31

Your PR is a backport of https://github.com/OCA/connector-telephony/commit/ccc142c02f5b7b041b28fb86fa6f585cd57cf6aa
Please include the comment in your backport.

pedrobaeza:
@oscars8a then you must cherry-pick the commit

oscars8a:
I will do it. Thank you so much for everything.

PR #204, opened 2019-12-13
[ADD] Migration of sales team on phone calls

StefanRijnhart:


pedrobaeza:
You should see https://github.com/OCA/crm/tree/9.0/crm_phonecall, as the module was duplicated in this version. There you have already the migration scripts.

StefanRijnhart:
OK, thanks for the pointer!

PR #209, opened 2020-03-27
[12.0][MIG] - base_sms_client migration to 12.0

sbejaoui:


sbejaoui:
cc/ @gva-acsone

cvinh:
@sbejaoui you removed the keychain dependance, where do you setup the login, password and sms_account with ovh ?

LuceEasyPme:
Hello @sbejaoui 
It would be good to have these 3 checks in wizards/mass_sms.py, on the send function : 
sms._check_gateway_method()
sms._check_gateway_permission()
sms._check_sms_length()

The problem that we have, is that we can click on "send" in the wizard, event thought the message is too long, and if we do not check the message list on technical, we don't know that the message was not actually sent.

Thanks,

cc @cvinh

LuceEasyPme:
Here is my PR on @sbejaoui repo, for the changes I suggested: 
https://github.com/acsone/connector-telephony/pull/1

cvinh:
Hello
Finally, we think that it's better to use the core sms module provided by Odoo CE since v12
We use it in production with the [ovh endpoint module](https://github.com/fmdl/odoo_apps_public) developped by @fmdl 
So we are not going to work anymore on that module
Best regards

sebastienbeau:
@cvinh @sbejaoui @LuceEasyPme maybe you can be interested by this PR https://github.com/OCA/connector-telephony/pull/217
I take a look to the @fmdl module but I prefers the api http2sms (really simple for the end user to configure and with several option on ovh interface quota, alert ...).

cvinh:
@sebastienbeau yes we like it too for same reasons, moreover, in OVH API there is a security hole as we can use the user account without password
Thank you !

dreispt (changes requested):
I'm sorry but the commit history is missing.
This needs to follow the migration instruction steps provided in the migration issue.

PR #211, opened 2020-04-01
12.0 mig ovh sms client

cvinh:


cvinh:
needs https://github.com/OCA/server-tools/pull/1802

bealdav:
FYI keychain is no more used in v12

see https://github.com/OCA/server-auth/pull/171#issuecomment-604980689

cvinh:
Hello
Finally, we think that it's better to use the core sms module provided by Odoo CE since v12
We use it in production with the [ovh endpoint module](https://github.com/fmdl/odoo_apps_public) developped by @fmdl
So we are not going to work anymore on that PR
Best regards

Issue #214, opened 2020-06-08
[12] base_phone (via asterisk_click2dial) : web client jumps to the default tab after any (un)successful call


luffah:
Hello,

  Module(s) : asterisk_click2dial (, base_phone)
  Version : 12.0.1.0.0 (https://github.com/OCA/connector-telephony/commit/6d9df1037a97dbf791a7f0a717846a63144174bb)

  when i use asterisk_click2dial  (dial button) on a contact sheet (model=res.partner&view_type=form),
 the Odoo web client acts like it was jumping to the first item in the main menu list (default tab).
 (Is it crash that cause this behavior ? i see exeptions in javascript console)

 I got this issue in 2 cases :
  - either the Asterisk server is well configured and the call is done
  - or the Asterisk server is not configured successfully and the call fails
 In both case, i use the functionality from a contact sheet, and after the call it goes back to the default tab.

## Step to reproduce 
 - (install module)
 - Set up your asterisk config in Odoo.
 - Set up a contact with a phone number.
 - Search contact in contact list.
 - Go to the contact form.
 - Click on  `☎ Dial` button. 
 - Close the popup (if needed) after call ends (or fails).

## What seems to be a normal behavior
 - after clicked on Dial button the current view on the contact shall stay in background;  when the call is ended it goes back to the contact form (origin of the call)

luffah:
I tried to remove module _asterisk_click2dial_ and forcing `true` to show dial button. When i click on it, it tries a dummy call and have the effect described in the issue.

luffah:
I didn't find in Odoo12 doc how to preserve location, therefore (as a workaround) i could use some dirty javascript tricks in base_phone js part :
```javascript
        click2dial: function(phone_num) {
            .
            .
            var orig_location = window.location.href;
            .
            .
            .
            window.location.href = orig_location;
            // window.location.reload(true); 
            .
            .
            .
```

See commit https://github.com/osiell/connector-telephony/commit/a2cb8d698e440473529516cba8df45d736138f16

filipocc:
> I didn't find in Odoo12 doc how to preserve location, therefore (as a workaround) i could use some dirty javascript tricks in base_phone js part :
> 
> ```js
>         click2dial: function(phone_num) {
>             .
>             .
>             var orig_location = window.location.href;
>             .
>             .
>             .
>             window.location.href = orig_location;
>             // window.location.reload(true); 
>             .
>             .
>             .
> ```
> 
[...]

PR #215, opened 2020-06-09
[IMP] base_phone : (refactoring) remove controller

luffah:
Hello, 

  here is just a proposal to simplify module structure by directly call `click2dial` function passing phone_number and context in the function call (js side), instead of using a controller for that.


alexis-via:
Seems a good idea. Did you test that code in real life with an asterisk server ?

luffah:
Implicitly : yes. (the -customer- testing team says me it is ok, and since i have no personal asterisk server ... i have no other choice to trust em)

They use this branch  https://github.com/osiell/connector-telephony/tree/12.0-back-to-orig (fixing https://github.com/OCA/connector-telephony/issues/214) based on the current proposed change, that i will propose once this one is (or not) merged.

alexis-via (approved):

alexis-via:
Thanks for your contribution !

PR #219, opened 2020-07-22
[12.0] base_phone : preserve location if no action is returned by the server

luffah:
Related to issue : #214

This PR add (javascript) redirection to the object (partner, sale ...) from which the dial button has been clicked.


luffah:
Tested ok by one of our customer's testing team.

Don't hesitate to redo testing ^^ (can be partially tested with a dummy sip server)

PR #223, opened 2020-09-17
[10.0] crm_phone : fixes in crm.phonecall view

alexis-via:
Fix display of icon on crm.phonecall tree view (convert old icon to fontawsome)
Give access to crm.phonecall form view (remove editable="top")


alexis-via:
Travis error is unrelated.

PR #251, opened 2021-07-05
[IMP] Updating state sms as sent when saved

Kev-Roche:
Change sms state as "sent" when they are sent and saved. Otherwise, the sms state keeps its default value : outgoing.

### DIFF
--- a/sms_no_automatic_delete/tests/test_purge_and_unlink_sms.py
+++ b/sms_no_automatic_delete/tests/test_purge_and_unlink_sms.py
@@ -5,29 +5,49 @@ from datetime import datetime, timedelta
 
 from odoo.tests.common import TransactionCase
 
+from odoo.addons.sms.tests.common import MockSMS
 
-class TestSmsPurge(TransactionCase):
+
+class TestSmsPurgeAndUnlink(TransactionCase, MockSMS):
     def setUp(self):
-        super(TestSmsPurge, self).setUp()
+        super(TestSmsPurgeAndUnlink, self).setUp()
         self.sms_sms = self.env["sms.sms"]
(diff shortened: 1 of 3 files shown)


Kev-Roche:
cc @sebastienbeau , @bguillot

sms_no_automatic_delete/tests/test_purge_and_unlink_sms.py
```diff
@@ -5,30 +5,50 @@
...
+        self.assertEqual(self.sms2.state, "outgoing")
+
+        with self.mockSMSGateway():
+            self.sms2.send()
+        self.assertEqual(self.sms2.state, "sent")
+
+    def test_sms_unlink(self):
+        # force_unlink=False by default
+        self.sms1.unlink()
+        self.assertEqual(len(self.sms_sms.search([])), 3)
```
sebastienbeau: do not do a search, as this test will have random result.
If a module create a sms and so you already have an sms the qty will be 4.

you can simply do

```
self.sms1.unlink()
self.assertTrue(self.sms1.exists())
self.sms1.with_context(force_unlink=True).unlink()
self.assertFalse(self.sms1.exists())
```

Kev-Roche: Indeed, it's better your way, I did the modification

sebastienbeau (approved):
LGTM (code review) please squash

Kev-Roche:
cc @sebastienbeau

sebastienbeau:
@Kev-Roche I think you did something wrong we should see only one commit here

PierrickBrun (approved):
LGTM

PR #252, opened 2021-07-18
[IMP] new module to send SMS using Clickatell

rpsjr:
This is a new module to use odoo base SMS module to send SMS using Clickatell gateway.

PR #255, opened 2021-07-25
[NEW] Send SMS with Clickatell API: new module sms_clickatell

rpsjr:
This is a new module to use odoo base SMS module to send SMS using Clickatell gateway.

remake of #252


marcelsavegnago:
@rpsjr Separate the readme file into files like the example below please.
https://github.com/OCA/connector-telephony/tree/12.0/sms_ovh_http/readme

Here is an example of how to generate the readme file. However, you don't need to worry about this because the oca bot already does this.
https://github.com/OCA/maintainer-tools#changelog-generator-using-towncrier

About PR #252 you can close it yourself.

PR #263, opened 2021-12-17
[14.0][ADD] sms_no_alter_body module

Kev-Roche:
When sending sms with the "_message_sms" method, some formatting are occuring between html and text. Consequently, if an URL is present in the initial sms's body, it will appeared twice in the final sms. This module avoids this kind of behavior.

### DIFF
--- a/sms_no_alter_body/models/mail_thread.py
+++ b/sms_no_alter_body/models/mail_thread.py
@@ -1,0 +1,14 @@
+# Copyright (C) 2021 Akretion (<http://www.akretion.com>).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    @api.returns("mail.message", lambda value: value.id)
+    def _message_sms(self, body, **kwargs):
+        self = self.with_context(force_sms_body=body)
+        return super()._message_sms(body, **kwargs)
--- a/sms_no_alter_body/models/sms_sms.py
+++ b/sms_no_alter_body/models/sms_sms.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2021 Akretion (<http://www.akretion.com>).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SmsSms(models.Model):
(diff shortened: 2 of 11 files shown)


sms_no_alter_body/models/mail_thread.py
```diff
@@ -0,0 +1,16 @@
...
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    @api.returns("mail.message", lambda value: value.id)
+    def _message_sms(self, body, **kwargs):
+        if self._context.get("sms_no_alter_body") is not False:
+            test = self.sudo().with_context(sms_no_alter_body=body)
+            return super(MailThread, test)._message_sms(body, **kwargs)
+        return super()._message_sms(body, **kwargs)
```
sebastienbeau: What we want is to always keep the body when posting a message
The method _message_sms is call in two place in odoo
Here: https://github.com/odoo/odoo/blob/5db838cfc91d8438d6234df00bd8174768445974/addons/sms/models/mail_thread.py#L193
And here : https://github.com/odoo/odoo/blob/5db838cfc91d8438d6234df00bd8174768445974/addons/sms/wizard/sms_composer.py#L243

In both case the body is text and there is no need to alter it, so I see no case where the alteration is needed, so never alter it.

```suggestion
    def _message_sms(self, body, **kwargs):
        self.with_context(force_sms_body=body)
        return super()._message_sms(body, **kwargs)
```

sms_no_alter_body/models/sms_sms.py
```diff
@@ -0,0 +1,17 @@
...
+class SmsSms(models.Model):
+    _inherit = "sms.sms"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        sms_no_alter = self._context.get("sms_no_alter_body")
+        if sms_no_alter:
+            for vals in vals_list:
+                vals["body"] = self._context["sms_no_alter_body"]
+        return super().create(vals_list)
```
sebastienbeau:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        for vals in vals_list:
        if self._context._get("force_sms_body"):
            for vals in vals_list:
                vals["body"] = self._context._get("force_sms_body")
        return super().create(vals_list)
```

sms_no_alter_body/models/mail_thread.py
Kev-Roche: You re right

sebastienbeau (approved):
Lgtm code review

Kev-Roche:
cc @kevinkhao

kevinkhao (approved):
code review LGTM

PR #286, opened 2022-12-21
[15.0][MIG] base_phone

Aldeigja:
Standard migration to 15.0


base_phone/i18n/de.po
```diff
@@ -46,7 +46,7 @@ msgstr "Click2dial erfolgreich"
 #, fuzzy, python-format
 #| msgid "Click2dial started"
 msgid "Click2dial to %s"
-msgstr "Click2dial gestartet"
+msgstr ""
```
GabbasovDinar, line 49: why are you delete ir?

GabbasovDinar:
LGTM!

GabbasovDinar (approved):

Aldeigja:
@OCA/connector-telephony-maintainers could you review?

hunghn (approved):

JorgeQuinteros:
@OCA/connector-telephony-maintainers It seems to be stuck in the runbot, could you please check it?

Aldeigja:
> @OCA/connector-telephony-maintainers It seems to be stuck in the runbot, could you please check it?

@JorgeQuinteros I've rebased it so now runboat is working

JorgeQuinteros:
Hi! Can anyone merge this PR please? its seem ready to merge.
CC: @Aldeigja

dantiller2:
Can this be merged in please?

base_phone/models/phone_validation_mixin.py
```diff
@@ -5,13 +5,28 @@
 
 from odoo import models
 
+from odoo.addons.phone_validation.tools import phone_validation
+
 
 class PhoneValidationMixin(models.AbstractModel):
-    _inherit = "phone.validation.mixin"
+    _name = "phone.validation.mixin"
```
pedrobaeza, line 12: This is not correct. You should inherit the ORM way. You can't replace a whole model.

pedrobaeza (changes requested):

base_phone/models/phone_validation_mixin.py
llacroix: @pedrobaeza the model has been removed in v15. So it can't be inherited.

llacroix: If I had to be nitpicky, I'd probably just want to use different mixin name or not at all. The functions can be called directly instead of having the methods into the models.

pedrobaeza: I see, but you should inherit the new mixin name instead (`mail.thread.phone`), isn't it?:

https://github.com/odoo/odoo/blob/15.0/addons/phone_validation/models/mail_thread_phone.py#L29

llacroix: I don't think so.

This isn't a new mixin, see here: https://github.com/odoo/odoo/blob/14.0/addons/phone_validation/models/mail_thread_phone.py (14.0) 

The main difference between 14.0 and 15.0 is that the mixin for validation has been effectively removed and replaced by methods in this file: 

https://github.com/odoo/odoo/blob/15.0/addons/phone_validation/tools/phone_validation.py

I don't really see any issue other than the mixin should probably get removed completely... On the other hand, it would be hard to make it customizable without the mixin. One would have to monkey patch the phone_validation.py file to extend functionality. That doesn't sound better than restoring the removed mixin.

ivs-cetmix: hi @pedrobaeza @llacroix any chances to have it resolved and merged?

PR #292, opened 2023-02-20
[14.0] asterisk_click2dial: AMI -> ARI transition

alexis-via:


alexis-via:
I wrote it several months ago, but kind of forgot to push it on github and make the PR. Hopefull, it was still on my hard-drive :)
Successfully tested today with a real asterisk server.

PR #294, opened 2023-02-22
[16.0][MIG] base_phone: Migration to 16.0

hunghn:
Based upon https://github.com/OCA/connector-telephony/pull/286


base_phone/models/phone_validation_mixin.py
```diff
@@ -5,13 +5,28 @@
...
+        country = country or self._phone_get_country()
+        if not country:
+            return number
+        return phone_validation.phone_format(
+            number,
+            country.code if country else None,
+            country.phone_code if country else None,
+            force_format="INTERNATIONAL",
+            raise_exception=False,
+        )
```
mmequignon, line 32: please use `odoo.addons.phone_validation.tools.phone_format`.
This makes this mixin kind of useless to me.

mmequignon: Therefore, I don't think this modules has any purpose anymore.

mmequignon (changes requested):

PR #287, opened 2022-12-21
[15.0][MIG] hr_phone

Aldeigja:
Standard migration to 15.0

PR #295, opened 2023-02-28
[16.0][MIG] hr_phone: Migration to 16.0

hunghn:
Based upon https://github.com/OCA/connector-telephony/pull/287
Depends on https://github.com/OCA/connector-telephony/pull/294


hr_phone/models/hr_employee.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2012-2021 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class HrEmployeePrivate(models.Model):
```
mmequignon, line 8: As I commented in the `base_phone` PR, I don't think this `phone.validation.mixing` makes sense anymore.
Use `odoo.addons.phone_validation.tools.phone_format` instead.

mmequignon (changes requested):

PR #303, opened 2023-08-11
[14.0][ADD] sms_twilio

mariadforgeflow:
Add Twilio as a IAP provider to send sms from Odoo


dnplkndll:
@JordiBForgeFlow review?

PR #300, opened 2023-05-16
[16.0][MIG] base_phone: Migration to 16.0

hailangvn2023:
* This is to replace https://github.com/OCA/connector-telephony/pull/296.

PR #314, opened 2024-02-15
[MIG] base_phone: Migrate to version 16.0

sonhd91:
This is to replace 
* https://github.com/OCA/connector-telephony/pull/300


etobella:
Sorry, We missed that you already did the migration and the other PR is already merged. :disappointed:

PR #327, opened 2024-12-27
[16.0][ADD] sms_alternative_provider, sms_messagebird

hbrunn:
This module allows configuring multiple providers for sending SMS.

By restricting providers to country or area codes, you can ie have Odoo pick different providers for different countries.


thomaspaulb:
@sebastienbeau This is an attempt to produce something a step up from what you have done with `iap_alternative_provider`, unfortunately the both cannot be reconciled. What do you think about it?

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -0,0 +1,183 @@
...
+        for this in self:
+            this.description = getattr(
+                this, "_get_description_%s" % this.gateway_type, lambda: False
+            )()
+
+    # SMS sending functions
+
+    def send(self, messages):
+        """
+        Send a list of SMS via the current provider
```
thomaspaulb, line 47: I'm not quite sure what I think about this function.

1. It's easily confused with `_send`, which does something decidedly different [how about `_send_via_self`?]
2. If you override `_send`, any call to `send` will bypass the overridden code. So where do you override if you want something done on all sends? [Think I got the answer to this already - `_send` calls `send`, so you'd override `send`]
3. Let's assume people actually start using this. Then when some neutralization script disables all providers, or leaves only one, any direct calls still use the disabled provider. (i had this once, with mails - could be solved by checking the active bit before sending)
4. I can't think of any concrete example in real life where I would want to choose the provider that I use from code. 99% of times people will have one provider, and in my case where I have several, I'd want this module to make the choice based on the type of message.

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -0,0 +1,183 @@
...
+import operator
+
+from odoo import _, api, exceptions, fields, models
+
+
+class IrSmsGateway(models.Model):
+
+    _name = "ir.sms.gateway"
+    _order = "sequence"
+    _desc = "SMS gateway provider"
```
thomaspaulb: This should be `_description`.

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -0,0 +1,183 @@
...
+        if not self.prefix:
+            return True
+        return any(
+            (message.get("number") or "").startswith(prefix)
+            for prefix in self.prefix.split()
+        )
+
+    def _handle_results(self, results, unlink_failed=False, unlink_sent=True):
+        """
+        Write state of sms.sms objects based on results
```
thomaspaulb: It might be nice to describe here which results are passed.

Also, are any of the original message key/value pairs passed? Would be handy to be able to rely on that.

```
            'id': sms.sms id,
            'number': phone number,
            'content': sms content,
```

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -0,0 +1,183 @@
...
+        providers = self._send_get_providers(messages)
+        provider2messages = providers._send_partition_providers(messages)
+        for provider, messages_to_send in provider2messages.items():
+            if not provider and raise_exception:
+                raise exceptions.UserError(
+                    _("No suitable provider found for messages %s") % messages_to_send
+                )
+            provider_result = provider.send(messages_to_send) or []
+            sms = SmsSms.browse(
+                filter(None, map(operator.itemgetter("id"), provider_result))
```
thomaspaulb: Would be nice here to catch any `KeyError` and then log a verbose message stating that the module implementor should have added `id` to the dict, before raising again; which is a bit clearer than `Sent batch 1 SMS: [16]: failed with exception 'id'`

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -0,0 +1,183 @@
...
+    @api.model
+    def _send(self, messages, handle_results=True, raise_exception=True):
+        """
+        Select provider(s) based on messages, call their provider specific function
+        _send_$gateway_type to actually send SMS
+
+        messages is a list of dictionaries [{
+            'id': sms.sms id,
+            'number': phone number,
+            'content': sms content,
```
thomaspaulb, line 68: Discussion point - some providers, not Odoo IAP I think but others, provide the possibility to specify a FROM number. Originally this can come either from a default specified on the account, or from the sms template. But in the end it has to end up in this dictionary. What would be the best way of supporting that?

sms_alternative_provider/models/ir_sms_gateway.py
hbrunn: agreed, made with private with your name proposal

sms_alternative_provider/models/ir_sms_gateway.py
hbrunn: I now pass the whole list of messages as passed to _send, seems most general to me

sms_alternative_provider/models/ir_sms_gateway.py
hbrunn: actually we have to support no it being passed as well for input as output, because just sending an sms to [some number](https://github.com/OCA/OCB/blob/16.0/addons/sms/wizard/sms_composer.py#L201) will pass 0 as id - so I made this `get` instead of `[]`

sms_alternative_provider/models/ir_sms_gateway.py
hbrunn: the default on the account is easy, as the provider-specific send method is called on the record of the account. The messagebird implementation I added here takes this from the company's name.

I don't see why we need this to be passed in this dict at all?

It will be difficult to do this on the template level, because in general an sms doesn't know which template it comes from (see the hack I'm using to solve this for sms_masked_content)

hbrunn:
another discussion point: Given that basically all modules that mess with sms at all override sms.api#_send_sms_batch and/or sms.sms#_postprocess_iap_sent_sms, it's crucial that the overrides of sms_alternative_provider come after any other override.

I'm of a mind to patch into loading the module graph and inject a dependency on sms_alternative_provider for every module that depends on sms if sms_alternative_provider is installed. Do you agree, or should we leave this to glue modules to be done on a per case basis?

thomaspaulb:
> it's crucial that the overrides of sms_alternative_provider come after any other override.

I'm not quite sure of this TBH: let's say someone wanted to use `sms_ovh_http` in its current form next to this module, they could even without the patch, because the OVH messages will just be caught earlier but when `super` comes, ours will be sent.

That's not a very likely scenario, but in other cases, if I would override `send_sms_batch` for other reasons and I call super before I do some stuff, I would want the SMS already be sent when the super returns; your proposal would mess with that.

Could you give an example use case where someone would want this?

sms_alternative_provider/models/ir_sms_gateway.py
thomaspaulb: Well, let's leave it for now. If I really need it I could always do a custom override to pass the sms template or the from-number as an extra provider-specific parameter and hack it in somehow. And then propose a generic fix, if needed. For now I'm setting it in my custom module via the sms account.

hbrunn:
sms_ovh_http works in both cases, so that's not a good example. sms_masked_content would have a problem, but that's handled accidentally right because modules on the same level (=same dependencies, both depend just on sms) are [sorted alphabetically](https://github.com/OCA/OCB/blob/16.0/odoo/modules/graph.py#L103).

So if you had another module also just depending on sms, but named `anything_sorted_before_sms_alternative_provider`, and it does something to messages that should affect all messages, currently this would only apply to messages sent via iap, because messages from other providers will be funneled away at this point already.

thomaspaulb:
Ah... right. Yes. Well your dependency idea does sound like a nice solution, so IMO you can go for it!

hbrunn:
ja hmm, i didn't think this through: at the point we can patch something, a module that we want to depend on ours might already be loaded, meaning its overrides are already incorporated into the models in the registry (odoo doesn't load the python code any more before actually loading the module). So we'd have to rebuild those models, remove the module from the module graph and have it reloaded afterwards.

that's a very heavy operation for a theoretical problem, so I suggest we stash this idea for now.

thomaspaulb (approved):
As for me I'm happy with the way it is now, we can work with it.

sms_alternative_provider/models/sms_sms.py
```diff
@@ -0,0 +1,41 @@
...
+                IrSmsGateway.browse(result.get("sms_gateway_id", [])), []
+            ).append(result)
+
+        actual_iap_results = []
+        for provider, results in provider2results.items():
+            if not provider or provider.gateway_type == "iap":
+                actual_iap_results.extend(results)
+            else:
+                provider._handle_results(
+                    [], results, unlink_failed=unlink_failed, unlink_sent=unlink_sent
```
thomaspaulb, line 32: Hey, the message dicts are not passed here,

sms_alternative_provider/models/sms_api.py
```diff
@@ -0,0 +1,20 @@
...
+    @api.model
+    def _send_sms_batch(self, messages):
+        """
+        Divert SMS sending to the gateway model if not asked not to
+        """
+        if self.env.context.get("force_iap"):
+            return super()._send_sms_batch(messages)
+        return self.env["ir.sms.gateway"]._send(
+            [dict(message, id=message.get("res_id")) for message in messages],
+            handle_results=False,
```
thomaspaulb, line 19: So in some cases, handle_results=False, which does not call handle_results directly after sending; instead it's being called from postprocess_iap_results, but then without the original message dicts.

What are the scenarios for having handle_results=True or False?

hbrunn: we want to support just saying

```
self.env["ir.sms.gateway"]._send([some dicts, may or may not link to sms.sms records])
```

in which case the send function should call `_handle_results`, and here it shouldn't because `_postprocess_iap_sent_sms` does it. So this clumsiness is caused by this module trying to introduce a new kind of sms api, while staying compatible with overrides other modules might do for the original api. We could consider dropping this.

sms_alternative_provider/models/sms_sms.py
hbrunn: there are none at this point, because this function is called in `sms.sms#_send`, which operates on `sms.sms` records, and those you can access via `results[i]['id']`

actually I don't understand any more why you requested having the message dicts passed here in the first place. If a provider wants to do provider-specific stuff, why wouldn't it do that in `_send_$provider` after sending? And note that whatever a provider wants to write on an sms.sms record, it can just add to its result dictionaries.

thomaspaulb:
Applying no stale since I think this `sms_alternative_provider` approach we use is important to be included, as the current `iap_alternative_provider` approach is limited. @OCA/connector-telephony-maintainers

ntsirintanis (approved):

NL66278 (approved):
👍 LGTM

hbrunn:
@thomaspaulb since https://github.com/OCA/OCB/commit/851f83fc5d6cbd70d5b801ceed54b46d38874afb the tests need another patch to succeed. Then the sms_ovh_http code only succeeds because it creates its own sms account, if that doesn't exist we run into https://github.com/OCA/OCB/blob/16.0/addons/iap/models/iap_account.py#L55 which returns an account from another transaction, causing accessing it afterwards to fail. So I think it's prudent anyways to test for existence there.

If you agree with the latest two fixup commits, please give me the chance to squash commits before merging

thomaspaulb:
@hbrunn Sorry for the delay here. I agree with the fixup commits, so if you squash, I will merge it.

hbrunn:
@thomaspaulb done

FranzPoize:
Hi !

I know I coming at the end but in odoo 18.0 there is a mechanism to allow for alternate provider (It is not well thought out but it exists and implements the logic just fine) using this https://github.com/odoo/odoo/blob/19.0/addons/sms/tools/sms_api.py and dispatching this sms_api depending on the res_company.

We talked about it with @sebastienbeau and we thought it would be a good idea to use this pathway to avoid reimplementing the mechanism implemented in https://github.com/odoo/odoo/blob/9501f8f97c8ff923eae52ece9b9390706fea63fe/addons/sms/models/sms_sms.py#L173C5-L173C83

What we propose is to keep the sms_alternative_provider but instead of hijacking the flow of sms_sms rewrite the way the api class is retrieved to avoid rewriting all the logic that is already in odoo.

What is your opinion on that ? If you're okay with this I could try to tweak your development to make it fit into with minimal change to the frontend the user experience.

Also this PR adds 3 modules wouldn't it be better to make 3 PR with 2 depending on this one ?

hbrunn:
for later Odoo versions, sure. But that's more an implementation detail of the migration to v17+, isn't it?

FranzPoize:
I'm so sorry I thought this was the migration to 18.0. I'm sorry ignore all this

hbrunn:
well still your migrating this to v18 will be very appreciated

thomaspaulb:
@FranzPoize Well at least with this merged then a regular forward-port including any proposed changed can be started the proper way.

PR #333, opened 2025-03-07
[ADD] voip_oca

etobella:
This module allows the use of VOIP directly from Odoo.

It relies on SIP.js to connect to the PBX using a websocket directly from the browser.

Odoo server will not connect directly to the PBX server, but will have users and passwords stored.

Credits
======

The development of this module has been financially supported by:

- Associacion Española de Odoo ([AEODOO](https://www.aeodoo.org/))

Configuration
=============

Create the PBX Connection
-------------------------

-  Access in Debug mode.
-  Go to ``Settings > Technical > Discuss > PBX Servers``.
-  Create a PBX server and define the domain name and websocket link.

You can set it as ``Development`` or ``Production``. Production environment will never contact the PBX server.

Configure users
---------------

For each user, we need to define their PBX server, user and password. To do this, we have two options:

1. Configure users as an admin and define the information directly 
2. As a user, go to your Preferences and modify the information.

Usage
=====

Automatically the system will login into the PBX server once we enter odoo with our user. If we are on a test environment, no login is made and the libraries of SIP are not loaded.

Once we are logged, we can use the VOIP widget in order to call directly. In order to open all the selectors, we can press on the top bar and the bottom widget will be opened.

image

We can make call by:

1. Select the partner and press the number
2. Select an activity and press the number of the associated partner
3. Select an old call and press the number
4. Press the keyboard and write the number directly

image

Inside a call, we can transfer the call, end the call, mute the call and put the call on Hold.

Also, the system allows to receive call. In that case, the system will try to find the related partner and will open the widget automatically.

image

Known issues / Roadmap
======================

The following options are interesting but are not implemented right now:

-  Allow to enable or disable the VOIP as a user (Login / Logout)
-  Create an autocall system where the calls are done automatically according to tasks

The following options will be done before release:

- [x]  Show the status of the VOIP connection on the header
- [x] Allow to call from a phone widget


voip_oca/README.rst
```diff
@@ -0,0 +1,171 @@
...
+=============
+
+Create the PBX Connection
+-------------------------
+
+-  Access in Debug mode.
+-  Go to ``Settings > Technical > Discuss > PBX Servers``.
+-  Create a PBX server and define the domain name and websocket link.
+
+You can set it as ``Development`` or ``Production``. Production
```
alexis-via: At the end of line: Production => Development

voip_oca/models/mail_activity.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.osv import expression
+
+
+class VoipOcaActivity(models.Model):
+    _inherit = "mail.activity"
+
+    main_partner_id = fields.Many2one(
```
JordiBForgeFlow, line 11: I would suggest to make the module dependent on mail_activity_partner https://github.com/OCA/social/pull/1606 instead.

pedrobaeza: Please don't. Base modules shouldn't depend on others not base one. I think the activity should be put to the user, not the partner.

etobella: But we need the partner to show the partner that we are going to call. That's why we used this.

The problem with the module that you are suggesting is that we are adding the extra mail_board functionality that it is not required right now for us :thinking:

voip_oca/static/src/components/call_list/call_list.xml
```diff
@@ -11,8 +11,8 @@
                 <span class="fw-bold" t-esc="call.partner[1]" />
                 <span t-esc="call.displayName" />
                 <div class="small text-muted">
-                    <t t-esc="call.startDate" />
-                    <t t-if="call.duration"> • <t t-esc="call.duration" />
+                    <t t-esc="call.dateStr" />
+                    <t t-if="call.duration"> • <t t-esc="call.durationStr" />
```
luisDIXMIT, line 15: maybe should be 't-if="call.durationStr"'

etobella: no need, as durationStr is computed from duration

voip_oca/static/src/services/voip_oca_service.esm.js
```diff
@@ -0,0 +1,182 @@
...
+    get partnerProps() {
+        return {
+            partner: this.partner || {},
+            activity: this.activity || {},
+            call: this.call || {},
+        };
+    }
+
+    /* Search functions */
+    async searchPartners(_search = "", offset = 0, limit = 13) {
```
luisDIXMIT, line 130: I think that limit parameter must be set to 0 in order to retrieve all contacts not only the first 13 and get access to the full contact list. WDYT?

luisDIXMIT (changes requested):

voip_oca/static/src/services/voip_oca_service.esm.js
etobella: I prefer to keep it simple, otherwise the widget will not be able to handle it if the Database is too big... Also, we should limit the values in the widget if possible (not only on the backend connection)

luisDIXMIT: Copy mate. I see your point of view

voip_oca/models/mail_activity_mixin.py
```diff
@@ -0,0 +1,25 @@
...
+    def _compute_activity_main_partner_id(self):
+        for record in self:
+            record.activity_main_partner_id = record.get_activity_main_partner_id()
+
+    def get_activity_main_partner_id(self):
+        if self._name == "res.partner":
+            return self
+        if "partner_id" in self._fields:
+            return self.partner_id
+        return self.env["res.partner"]
```
pedrobaeza: No need of 2 methods. For inheritance, just override the compute:

```suggestion
    def _compute_activity_main_partner_id(self):
        self.activity_main_partner_id = False
        if "partner_id" in self._fields:
            for record in self:
                record.activity_main_partner_id = record.partner_id
```

and then on res.partner:

```
    def _compute_activity_main_partner_id(self):
        for record in self:
            record.activity_main_partner_id = record
```

voip_oca/models/res_partner.py
```diff
@@ -0,0 +1,32 @@
...
+            "type": "partner",
+            "displayName": self.display_name,
+            "email": self.email,
+            "landlineNumber": self.phone,
+            "mobileNumber": self.mobile,
+            "name": self.name,
+        }
+
+    @api.model
+    def get_voip_contacts(self, _search, offset, limit):
```
pedrobaeza: As I see you are doing some kind of "API", please prefix all the methods with `voip` / `_voip` instead:

```suggestion
    def voip_get_contacts(self, _search, offset, limit):
```

voip_oca/models/res_users.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @property
+    def SELF_WRITEABLE_FIELDS(self):
+        return super().SELF_WRITEABLE_FIELDS + [
+            "voip_pbx_id",
+            "voip_username",
+            "voip_password",
+        ]
+
+    def _get_voip_info(self):
```
pedrobaeza:
```suggestion
    def _voip_get_info(self):
```

voip_oca/models/voip_oca_call.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2025 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import api, fields, models
+from odoo.osv import expression
+
+
+class VoipOcaCall(models.Model):
+    _name = "voip.oca.call"
```
pedrobaeza: I don't think we need `oca` on the model (and if Odoo is calling this the same, it's not a problem, as both modules are incompatible between them and we shouldn't look for the compatibility IMO).

```suggestion
    _name = "voip.call"
```

voip_oca/models/voip_oca_pbx.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class VoipOcaPbx(models.Model):
+    _name = "voip.oca.pbx"
```
pedrobaeza:
```suggestion
    _name = "voip.pbx"
```

voip_oca/static/audio/calltone.mp3
pedrobaeza, line 1: Why not putting this in OGG?

voip_oca/static/description/actions.png
pedrobaeza, line 1: I think this should go into `static/src/images` and use relative image path in the .md

voip_oca/models/mail_activity_mixin.py
etobella: It is done this way in order to leave a hook for other modules.

pedrobaeza: As said, the hook can be done in the compute method in the mixin or in the specific model. No need of extra method.

voip_oca/static/audio/calltone.mp3
etobella: Because I found it by MP3. I can try to find OGG

voip_oca/static/description/actions.png
etobella: I don't understand it. It is used as relative paths now :thinking:

voip_oca/models/mail_activity_mixin.py
etobella: Yes, but it is easier to extend it this way. Also, there are examples of this behavior inside OCA and Odoo. I would prefer to keep it as it is

pedrobaeza: I don't think so, as the stacktrace is longer, and the code is less readable being split and this method is only called in one place, which is the clear indicator that shouldn't be extracted.

pedrobaeza:
Please check CI.

voip_oca/models/mail_activity_mixin.py
etobella: Well, it was done this way to have it done as it is right now (I forgot to split it correctly). With this,  we can change it by thread model. It is longer, yes, but it is cleaner too IMO. Also, we can find several examples in Odoo Core:

https://github.com/odoo/odoo/blob/c315f5af7bbe47d6fa73d0dd0ba31ab6f93cfd03/addons/mail/models/mail_thread.py#L429
https://github.com/odoo/odoo/blob/c315f5af7bbe47d6fa73d0dd0ba31ab6f93cfd03/addons/hr_recruitment/models/hr_job.py#L339

pedrobaeza: Well, that examples are not exactly the same and Odoo core can't be sometimes the best example, but this is not so important to block the PR.

nimarosa (approved):

luisDIXMIT (approved):
LGTM!!

pedrobaeza (approved):
According live tests.

/ocabot merge nobump

PR #334, opened 2025-03-09
[17.0][FIX] base_phone: Fix due to changes on Odoo

etobella:
Necessary to change function name due to 

https://github.com/odoo/odoo/commit/a823c4ef6962672fe68726682e94ff82db211772

### DIFF
--- a/base_phone/models/phone_validation_mixin.py
+++ b/base_phone/models/phone_validation_mixin.py
@@ -12,7 +12,7 @@ class PhoneValidationMixin(models.AbstractModel):
     _name = "phone.validation.mixin"
     _description = "Phone Validation Mixin"
 
-    def _phone_get_country(self):
+    def _phone_validation_get_country(self):
         if "country_id" in self and self.country_id:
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
I don't get the change,Enric. The pointed commit is adding `_phone_get_country` method, not `_phone_validation_get_country` one.

etobella:
Odoo added a function called  _phone_get_country 

https://github.com/odoo/odoo/commit/a823c4ef6962672fe68726682e94ff82db211772#diff-b59b1f7bd3ccab92fc11c0383009013d31d592305a7c5d8b1d5ecf6a7e508371R23-R27

And we had the same function name.

luisDIXMIT (approved):
I guess both functions have different logic, so both are necessary, and a rename is also needed to avoid conflicts. LGTM!!

pedrobaeza (approved):
OK, now I understand. I though it was an override. 

/ocabot merge patch

PR #336, opened 2025-03-24
[18.0][MIG] sms_no_automatic_delete: Migration to 18.0

chaule97:
- There is an autovacuum `def _gc_device(self)` which deletes the records every day (without calling `unlink` method). So, the cron `SMS: SMS Purge` is helpless now: 
  +  Add a new configuration "SMS Purge Days" : default = 90 (days)
  + Override `def _gc_device(self)` to delete the sms longer the "SMS Purge Days"


wlin-kencove (approved):

FranzPoize:
Hi @chaule97 could you recreate the PR ?

chaule97:
Hi @FranzPoize , Can you check https://github.com/OCA/connector-telephony/pull/362?

PR #345, opened 2025-08-01
[18.0][ADD] voip_oca_call_record: Module to record the audio of calls [WIP]

JordiBForgeFlow:
This module allows to record the audio in calls received and issued.

TO DO:

* [x] Define in the PBX when should calls be recorded (always/sample/at request by user)
* [ ] Add tests
* [x] Improve UI


fr33co:
I tested the module and it records well, but on one side only, only to the client.

PR #346, opened 2025-08-05
[17.0][FIX] voip_oca: Add compute_sudo to fields for proper access control

carlos-lopez-tecnativa:
In a multi-company environment, when Odoo searches for activities to display in the systray, an error is raised: "Uh-oh! Looks like you have stumbled upon some top-secret records." Using `sudo` prevents this error.

**Steps to reproduce:**

- In a fresh database with demo data, install the `voip_oca` and `account` modules.
- Create a new company.
- Switch to the new company.

The screen will be blank, and the menus will not load. Check the logs, and you will see the error.

image

### DIFF
--- a/voip_oca/models/mail_activity.py
+++ b/voip_oca/models/mail_activity.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
+from odoo.exceptions import AccessError
 from odoo.osv import expression
 
 
@@ -40,5 +41,16 @@ class VoipOcaActivity(models.Model):
                 [[(field, "ilike", _search)] for field in search_fields]
             )
             domain = expression.AND([domain, search_domain])
-        activities = self.search(domain, offset=offset, limit=limit)
-        return activities.activity_format()
+        all_activities = self.search(domain, offset=offset, limit=limit)
+        # Filter activities to avoid accessing records that the user cannot read
+        # due to multi-company restrictions or other access rules in the res_model.
+        no_access_ids = []
+        for activity in all_activities:
+            try:
+                res_record = self.env[activity.res_model].browse(activity.res_id)
+                res_record.check_access_rule("read")
+            except AccessError:
+                no_access_ids.append(activity.id)
+        if no_access_ids:
+            all_activities -= self.browse(no_access_ids)
+        return all_activities.activity_format()


etobella:
Mmmh, maybe we shouldn't see the activity, isn't it?

carlos-lopez-tecnativa:
> Mmmh, maybe we shouldn't see the activity, isn't it?

In the systray, the activities are displayed correctly according to the company. The issue happens in the computed field.
Do you think this error could be solved in another way?
Please let me know if you have any other suggestions.

Company 1
image

Company 2
image

etobella:
Well, I am not 100% of the solution,

However, my question is: IF the problem is related to the activity being shown when it shouldn't, maybe we should solve that, WDYT?

carlos-lopez-tecnativa:
I investigated a bit more, and the first function that searches for activities is: https://github.com/OCA/connector-telephony/blob/5be556f619da451e5a3ad96ea42e7a9f6735d720/voip_oca/models/mail_activity.py#L32-L43
So, I applied the filters on the related records.
Please let me know if this approach is better.

Sorry, I thought this issue was in the systray, but now I see the error is in the Activities view of VoIP.
image

etobella (approved):
Thanks!! Don't worry, for this reason we have reviews 😄 

No problem if all of us are happy with the outcome.

voip_oca/models/mail_activity.py
```diff
@@ -40,5 +41,15 @@ def get_call_activities(self, _search, offset, limit):
...
-        activities = self.search(domain, offset=offset, limit=limit)
+        all_activities = self.search(domain, offset=offset, limit=limit)
+        activities = self.env["mail.activity"]
+        # Filter activities to avoid accessing records that the user cannot read
+        # due to multi-company restrictions or other access rules in the res_model.
+        for activity in all_activities:
+            try:
+                res_record = self.env[activity.res_model].browse(activity.res_id)
+                res_record.check_access_rule("read")
+                activities |= activity
```
pedrobaeza: This is very slow. You should do other thing, like removing those that are not valid.

carlos-lopez-tecnativa: I updated the code. I tried to filter all problematic records and discard them from the activities if they exist.
What do you think?

pedrobaeza (approved):
That's better, thanks.

/ocabot merge patch

PR #347, opened 2025-08-05
[17.0][IMP] voip_oca: Update z-index for proper layering

carlos-lopez-tecnativa:
When the `web_responsive` module is installed, the VoIP component is not displayed in the main menu because `web_responsive` has a [higher z-index](https://github.com/OCA/web/blob/32184c6f6b6c34a2660e4adbe8271f2021f22bdf/web_responsive/static/src/components/apps_menu/apps_menu.scss#L35 ).

Before

https://github.com/user-attachments/assets/cbbedd60-b6e8-4be2-869a-e84047439386

After this commit, the VoIP component will always be displayed.

https://github.com/user-attachments/assets/f1ba597d-5e7e-4edc-8da8-e69eda84db7b

### DIFF
--- a/voip_oca/static/src/components/softphone/softphone.scss
+++ b/voip_oca/static/src/components/softphone/softphone.scss
@@ -2,7 +2,7 @@
     width: 300px;
     height: auto;
     max-height: 70%;
-    z-index: 99;
+    z-index: 1001;
 
     .o_voip_softphone_header {
         background-color: $o-community-color !important;


pedrobaeza:
At the same time you are there, can you please hide the systray icon when no voip provider is set in the user?

voip_oca/static/src/components/softphone/softphone.scss
```diff
@@ -2,7 +2,7 @@
     width: 300px;
     height: auto;
     max-height: 70%;
-    z-index: 99;
+    z-index: 1001;
```
pedrobaeza, line 5: I would say this z-index is on purpose for not being on top on certain cases, isn't it @etobella ?

pedrobaeza: Ping @etobella

etobella:
Could you try to show what happens if we open a readonly O2M line or opening a wizard?

pedrobaeza:
ping @carlos-lopez-tecnativa

carlos-lopez-tecnativa:
> Could you try to show what happens if we open a readonly O2M line or opening a wizard?


Opening a wizard
image

Opening an O2M (in the sale order, for example, I manually edited it to remove the editable attribute so I could open the line)
image

pedrobaeza (approved):

eduezerouali-tecnativa (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to 18

PR #352, opened 2025-10-13
[18.0][FIX] voip_oca: filter activities to avoid accessing records the user cannot read

carlos-lopez-tecnativa:
In a multi-company environment, when Odoo searches for activities to display in the systray, an error is raised:
"Uh-oh! Looks like you have stumbled upon some top-secret records."

This occurs because some activities point to records that the current user does not have access to, due to multi-company restrictions or other access rules defined in the `res_model`.

Steps to reproduce:

- In a fresh database with demo data, install the `voip_oca` and `account` modules.
- Create a new company.
- Switch to the new company.

Forward-port of https://github.com/OCA/connector-telephony/pull/346

### DIFF
--- a/voip_oca/models/mail_activity.py
+++ b/voip_oca/models/mail_activity.py
@@ -40,5 +40,12 @@ class VoipOcaActivity(models.Model):
                 [[(field, "ilike", _search)] for field in search_fields]
             )
             domain = expression.AND([domain, search_domain])
-        activities = self.search(domain, offset=offset, limit=limit)
-        return activities.activity_format()
+        all_activities = self.search(domain, offset=offset, limit=limit)
+        # Filter activities to avoid accessing records that the user cannot read
+        # due to multi-company restrictions or other access rules in the res_model.
+        allowed_activity_ids = []
+        for activity in all_activities:
+            res_record = self.env[activity.res_model].browse(activity.res_id)
+            if res_record._filtered_access("read"):
+                allowed_activity_ids.append(activity.id)
+        return self.browse(allowed_activity_ids).activity_format()


pedrobaeza (commented):
Isn't possible to reuse some code from upstream doing this? I would say Odoo core already performs something similar in the clock icon.

carlos-lopez-tecnativa:
> Isn't possible to reuse some code from upstream doing this? I would say Odoo core already performs something similar in the clock icon.

Odoo performs similar logic in this function, which is called from the Systray. However, this function cannot be reused here:
https://github.com/odoo/odoo/blob/44c993129241313c2011c6fff39478cc6791ae0f/addons/mail/models/res_users.py#L336-L343

This function uses `_filtered_access`, which internally calls `_check_access`. Therefore, it’s effectively the same logic as the code used here.

pedrobaeza (commented):
And why don't you use `_filtered_access`? I don't like too much the `try/except` code as it's less clean.

carlos-lopez-tecnativa:
> And why don't you use `_filtered_access`? I don't like too much the `try/except` code as it's less clean.

Done

pedrobaeza (approved):

voip_oca/models/mail_activity.py
```diff
@@ -40,5 +40,14 @@ def get_call_activities(self, _search, offset, limit):
                 [[(field, "ilike", _search)] for field in search_fields]
             )
             domain = expression.AND([domain, search_domain])
-        activities = self.search(domain, offset=offset, limit=limit)
-        return activities.activity_format()
+        all_activities = self.search(domain, offset=offset, limit=limit)
+        # Filter activities to avoid accessing records that the user cannot read
+        # due to multi-company restrictions or other access rules in the res_model.
+        no_access_ids = []
+        for activity in all_activities:
```
etobella, line 47: Wouldn't have sense to do access_ids y all_activities = self.browse(access_ids). Lo digo por que es un poco más sencillo de entender y con menos operaciones, no?

carlos-lopez-tecnativa: I have updated the code. Please let me know if it’s okay for you and if I understood what you meant.

carlos-lopez-tecnativa: Now I remember @pedrobaeza’s suggestion in this comment https://github.com/OCA/connector-telephony/pull/346#discussion_r2254528649
.
I’m not sure if this approach is better than the previous one. Please confirm.

etobella: Well, I think that both should work and give the same result, but you simplified it a bit. For me seems ok.

etobella (approved):

voip_oca/models/mail_activity.py
```diff
@@ -40,5 +40,12 @@ def get_call_activities(self, _search, offset, limit):
                 [[(field, "ilike", _search)] for field in search_fields]
             )
             domain = expression.AND([domain, search_domain])
-        activities = self.search(domain, offset=offset, limit=limit)
-        return activities.activity_format()
+        all_activities = self.search(domain, offset=offset, limit=limit)
+        # Filter activities to avoid accessing records that the user cannot read
+        # due to multi-company restrictions or other access rules in the res_model.
+        allowed_activities = []
```
pedrobaeza: Please name variable properly:

```suggestion
        allowed_activity_ids = []
```

PR #355, opened 2025-11-14
[Mig] sms_alternative_provider 

bjouini-acsone:
this is a partial migration of the Pr:https://github.com/OCA/connector-telephony/pull/327 on v16

the major changements from v16 to 18:
the sms_api model does not exist anymore 
on the previous implementation , if we found None passede we use the custom logic that we add , but in v18 if we pass None to the send method we always use the standard behaviour of odoo , this is why it is changed in this version


sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -37,94 +39,158 @@ def _compute_description(self):
...
+        Send SMS records via the current provider
 
         Return list of dictionaries [{
-            'id': sms.sms id,
-            'state': sms.sms#state,
-            'failure_type': sms.sms#failure_type
+            'uuid': sms.sms uuid,
+            'state': sms state,
+            'failure_reason': optional failure reason
         }]
```
qgroulard, line 54: We still return "id" as well.

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -37,94 +39,158 @@ def _compute_description(self):
...
     @api.model
     def _send_get_providers(self, messages):
         """
         Return all providers potentially suitable for the current context
         """
-        return self.search(
-            ["|", ("company_id", "=", False), ("company_id", "=", self.env.company.id)]
+        providers = self.search(
+            [
+                ("active", "=", True),
```
qgroulard: I don't think adding `("active", "=", True)` has any effect here.

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -37,94 +39,158 @@ def _compute_description(self):
...
+        for gateway, messages_to_send in gateway_groups.items():
+            if not gateway:
+                if raise_exception:
+                    raise exceptions.UserError(
+                        _("No gateway assigned for messages %s") % messages_to_send
+                    )
+                continue
+
+            sms_records = SmsSms.browse([msg["id"] for msg in messages_to_send])
+            provider_result = gateway._send_via_self(sms_records) or []
```
qgroulard: Could you please explain why you need to browse the sms record before calling `_send_via_self()` now ?

sms_alternative_provider/models/ir_sms_gateway.py
```diff
@@ -37,94 +39,158 @@ def _compute_description(self):
...
+                        "Assigned message %s to provider %s",
+                        message.get("id"),
+                        provider.name,
+                    )
                     break
-            else:
+            if not assigned:
                 result.setdefault(self.browse([]), []).append(message)
+                _logger.info("No provider found for message %s", message.get("id"))
         return result
```
qgroulard: In general, could you please try to minimize the diff by avoiding unnecessary changes.

For instance here I don't understand the change in the docstring.
Also the model `ir.sms.gateway` is ordered by sequence, thus I don't understand the sort you are adding.
Finally, how is different the use of your variable `assigned` from the previous `for else` ?

qgroulard (commented):
Hi, this is an attempt at reviewing this PR but:
1. I'm not really comfortable with this topic
2. I think it's lacking a lot of explanation

To ease the review, could you please:
1. Squash your 2 migration commits
2. Minimize the diff as much as possible, i.e. push only relevant changes. Please don't remove still accurate docstrings, reword working code, etc. Unless it really improves the code of course.
3. Provide explanation on the key changes. Please make use of the "description" section of this PR. Such a big PR should be heavily documented, not pushed with an empty description.

Once this PR is nice and clean, hopefully we can get the help of @hbrunn to get it reviewed and merged :+1:

yvaucher:
Overseeded by #360

PR #358, opened 2025-12-02
[18.0] [IMP] voip_oca: change rtcpMuxPolicy from 'require' to 'negotiate'

GiuTav:


etobella:
Justify the change, I am not sure if this should be modified, as this is the external library

GiuTav:
> Justify the change, I am not sure if this should be modified, as this is the external library

The change is justified because this parameter is not required for all PBX systems. By making it negotiated based on the PBX, the parameter is used only when necessary, depending on the specific configuration.

etobella:
But this is not the way for solving it. This comes from the library. You should try to find how to configure it and allow user to define it.

GiuTav:
> But this is not the way for solving it. This comes from the library. You should try to find how to configure it and allow user to define it.

I took inspiration from the Odoo Enterprise VoIP module, where the driver is configured so it can work with different PBX systems. With this approach, the parameter comes from the library but is exposed as a configuration option, so each user can define it according to their PBX instead of being constrained to a single fixed behavior

etobella:
You enforced the change on the library file, If you want to do that, you need to add an option on the pbx configuration and pass it.

GiuTav:
I would appreciate your advice on how to implement this change correctly. If you could guide me on adding the option to the PBX configuration and passing it properly, I would be very grateful.

etobella:
Add a field in voip.pbx with the possible values in voip.pbx and default value `require`.

Pass the value to the frontent using _voip_get_info in res.users

Add the parameter [here](https://github.com/OCA/connector-telephony/blob/18.0/voip_oca/static/src/services/voip_agent_service.esm.js#L36) using a new key
```
sessionDescriptionHandlerFactoryOptions: {
rtcpMuxPolicy: parameter,
}
```

Just warning that this is an approach, review it and improve it

PR #356, opened 2025-11-18
[18.0][MIG] sms_ovh_http: Migration to 18.0

FranzPoize:

PR #360, opened 2026-01-21
[18.0][MIG] sms_alternative_provider - use of SmsApiBase

anthonissen-a:
Use the new abstract class SmsApiBase from Odoo as a base to implement for any alternative provider.

The prefix isn’t implemented yet. Wouldn’t it be better to handle it in a separate module?


cvinh:
Thanks for this PR
Did you check the pending https://github.com/OCA/connector-telephony/pull/355
Could you comment why you did another PR or if you did not see it, maybe you could collaborate on it

anthonissen-a:
Hi @cvinh, yes this PR is based on #355. 
I did another PR because it suggests an alternative implementation of the solution, that uses the new SmsApiBase class of Odoo.

yvaucher:
@cvinh 

@anthonissen-a is taking over his teammate @bjouini-acsone  work made in #355 

I'll close #355

yvaucher:
I don't see the fixups in the history of the module.
https://github.com/OCA/connector-telephony/commits/16.0/sms_alternative_provider

From what I understand the history of this PR is based on the work in progress of #327 before it was squashed and merged.

Could you rebase this PR on top of a cherry pick of this commit feb05cf352e45e85776d45d20c49d07f2fda44e9

FranzPoize:
Hello !

Due to the fact that the odoo `sms_api` is hardwired to get the sms default `iap.account` for the sms service and that there is no way to simply get another one. I think we should add an `iap_account_id` field to the `ir.sms.gateway` or at least a mechanism to give an `iap.account` to the sms_api.

Something of the sort:
```python
# In sms_sms.py
    def _split_by_api(self):
        """Overwrite to select the default SMS API configured in SMS Gateway."""
        if not self.sms_gateway_id:
            self.sms_gateway_id = self.env["ir.sms.gateway"]._get_default_gateway()
        sms_api = self.sms_gateway_id._get_api_class()
        
        # This line is changed !
        yield sms_api(self.env, self.sms_gateway_id.iap_account_id), self

# In ir_sms_gateway.py
    iap_account_id = fields.Many2one("iap.account")
```

If you found a workaround please put an example somewhere in the module

FranzPoize:
Bump @anthonissen-a

anthonissen-a:
Hi @FranzPoize,

Thanks for your comment ! In fact it's a good idea, I will adapt the code with your suggestion.

FranzPoize (approved):
Test are passing and I am able to send sms using sms_ovh_http
LGTM

FranzPoize:
Can you take a look @sebastienbeau @hbrunn ?

sms_alternative_provider/models/sms_api.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo.addons.sms.tools.sms_api import (
+    SmsApi as OdooSmsApi,
+)
+from odoo.addons.sms.tools.sms_api import (
+    SmsApiBase as OdooSmsApiBase,
+)
+
+
+class SmsApiBase(OdooSmsApiBase):
```
hbrunn, line 13: how does this construction improve anything over implementations extending the gateway model? they will have to touch it anyways, and to me it seems much more straightforward to implement provider specific functionality there.

I'd imagine sms.sms#_split_by_api returning ir.sms.gateway instances, and ir.sms.gateway#_send_sms_batch calls getattr(self, f'_send_sms_batch_{self.gateway_type}'), pretty similar to what the original does

hbrunn (changes requested):

sms_alternative_provider/models/sms_api.py
FranzPoize: This allows the module to use the scaffolding given by the new sms architecture in odoo, the sms api was completely changed in 18.0 I think it's better to reflect that.

Why not adhere to what odoo provides and make a call to getattr with an obfuscated function call ?

I use this architecture in my migration of sms_ovh_http and it's pretty easy to inherit from this module as is.

And no we don't have to touch the gateway model to use a new provider (you can look at https://github.com/OCA/connector-telephony/pull/356 to see how)

hbrunn:
it's a poor choice to have the provider-specific implementations in a non-orm class because that makes overriding them a pain.

I see the type now actually lives in the iap.account model, so you have to touch that, doesn't matter for my point.

FranzPoize:
You don't need to override it more than once since if you want to use a provider, the API you call in your new ProviderSmsApi either works or don't work.

If it doesn't work inheriting from it doesn't really matter, if it work you don't need to inherit from it, if it's incomplete the module providing the ProviderSmsApi should be improved.

If you want to modify how the sms data is treated you can do it around the SmsApi calls.

If we don't use the SmsApi base class we have to rewrite it here I don't see how this is better.

PR #363, opened 2026-03-03
[18.0][FIX] base_phone: fix use of callerid in context

alexis-via:


nibraun (approved):
Successfully tested on a real Asterisk IPBX, with light code review

PR #364, opened 2026-03-04
[19.0][MIG] voip_oca

etobella:
Migration of the code.

Adapt to the new mail reality.


luisDIXMIT:
@etobella I fixed the tests in a separate commit. I can squash it if everything looks correct.

luisDIXMIT (commented):
I got this error when I tried to make a call. It might be my connection; it isn't configured correctly. I tested the module and everything else works fine.
image

dnplkndll (approved):
LGTM 👍 Tested locally 

  Two tiny nits, **neither blocking**, fix in a follow-up if you'd rather just ship this:

  **`voip_oca/static/description/icon.png`** is the OCA placeholder (8892 bytes). Delete
  it so `/ocabot merge` regenerates from the manifest.

  **`voip_oca/models/voip_call.py:44`** — `_compute_display_name` reads
  `rec.partner_id.display_name` and `rec.phone_number` (the empty-partner fallback), but
  `@api.depends` only declares `("state", "partner_id.name")`. The `partner_id.name`
  declaration cascades for name-only changes via Odoo's compute graph, but misses:
  - `phone_number` write on a partner-less call → `display_name` stays stale
  - `parent_id` rename on a child partner → ditto

  Suggested:
  ```python
  @api.depends("state", "partner_id.display_name", "phone_number")
```

Thanks for the migration!

pedrobaeza (approved):
/ocabot merge nobump

PR #371, opened 2026-07-01
[16.0][IMP] sms_ovh_http: migrate to OVH new API (python-ovh)

khoivo1601:
- Replace the legacy `http2sms.cgi` endpoint with the official OVH API using the `python-ovh` SDK


khoivo1601:
the `pre-commit` failure seems to be an environment issue in the CI config

